Issue #3, opened 2017-02-09
Merge patches from the old repository


ForNeVeR:
There's a bunch of patches attached to [the old repository tracker](https://bugs.launchpad.net/wpf-math). I think it should be worth to check them out and apply to our codebase.

- [x] [Feature: added color](https://bugs.launchpad.net/wpf-math/+bug/1052493)
- [x] [Rendering of accent symbols](https://bugs.launchpad.net/wpf-math/+bug/1033830)
- [x] [incorrect image size when using subscript](https://bugs.launchpad.net/wpf-math/+bug/1041219)
- [x] [Background isn't drawn when using a single char](https://bugs.launchpad.net/wpf-math/+bug/1046368)
- [x] [Image size: Italic support](https://bugs.launchpad.net/wpf-math/+bug/1046669)
- [x] [Incorrect XML serialization in different regions](https://bugs.launchpad.net/wpf-math/+bug/1050912)
- [x] [Feature: Render geometry & output as SVG](https://bugs.launchpad.net/wpf-math/+bug/1050955)
- [x] [Loading of predefined formulas](https://bugs.launchpad.net/wpf-math/+bug/1052586)
- [x] [Added whitespace and % rendering](https://bugs.launchpad.net/wpf-math/+bug/1052891)

Ideally we should treat these patches as our own pull requests: we should try our best to reproduce the mentioned issues (probably even write the unit tests), and merge the changes afterwards. The maintainer on duty (well, probably myself) should notify the patch authors that the patches are being merged here.

ForNeVeR:
Thanks to @gsomix, we've merged all the patches. I've notified all the possibly interested parties about that.

PR #8, opened 2017-02-11
Apply patches Rendering of accent symbols and Loading of predefined formulas

gsomix:
Thanks to Marc Palmans.

[Rendering of accent symbols](https://bugs.launchpad.net/wpf-math/+bug/1033830)

> Accent symbols are rendered separately instead of above other characters.
> The following patch appears to fix this issue but someone with more knowledge of the source code and latex should check it.

[Loading of predefined formulas](https://bugs.launchpad.net/wpf-math/+bug/1052586)

> I fixed some issues regarding the loading of predefinedformulas.xml Everything should work except \left and \right, but I think we should use TexFormulaHelper.AddEmbraced to do this. I've temporarily disabled those in the xml file.

## Example:

before patches
hbar before

after patches
hbar after

#3

### DIFF
--- a/WpfMath/Data/PredefinedTexFormulas.xml
+++ b/WpfMath/Data/PredefinedTexFormulas.xml
@@ -20,7 +20,7 @@
 
   <Formula name="hbar" enabled="true">
     <CreateFormula name="f">
-      <Argument type="string" value="\bar" />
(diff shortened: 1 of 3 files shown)


ForNeVeR:
Thanks both to you and to Marc Palmans for the patches.

PR #9, opened 2017-02-11
Apply patch Feature: added color

gsomix:
Another awesome patch by Marc Palmans.

[Feature: added color ](https://bugs.launchpad.net/wpf-math/+bug/1052493)
> I've added experimental code to render \color and \colorbox commands. I've also added an xml file with predefined colors. There is no support to add custom colors at the moment. This patch also fixes the foreground color in HorizontalRule which defaulted to the color black.

## Example:

\color
color

\colorbox
colorbox

#3


ForNeVeR:
Thanks for the patch!

PR #11, opened 2017-02-11
Apply patch Feature: Render geometry & output as SVG

gsomix:
Kudos to Marc Palmans!

[Feature: Render geometry & output as SVG](https://bugs.launchpad.net/wpf-math/+bug/1050955)
> I've added code to render to a Geometry and then convert it to an SVG snippet which then can be inserted in an SVG document. At the moment it's hard coded to black with no stroke as I don't need anything else. It should be very easy to add those features if you need them.

## Example:

render in Inkscape
svg-result

[svg-example.zip](https://github.com/ForNeVeR/wpf-math/files/768608/svg-example.zip)

#3

### DIFF
--- a/WpfMath.Example/MainWindow.xaml
+++ b/WpfMath.Example/MainWindow.xaml
@@ -24,8 +24,15 @@
 
         <DockPanel Margin="10" Grid.Row="0" Grid.Column="0">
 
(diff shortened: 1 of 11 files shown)


ForNeVeR:
Thanks!

PR #12, opened 2017-02-11
Apply patch Background isn't drawn when using a single char

gsomix:
Kudos to Marc Palmans.

[Background isn't drawn when using a single char](https://bugs.launchpad.net/wpf-math/+bug/1046368)
> See title.

## Example:

before patch
background before

after patch
background after

#3


ForNeVeR:
Thanks. Fixed the conflicts and rebased as aa875fa4557fa9c0f11f8df447291ef1959837b4.

PR #16, opened 2017-02-12
Apply patches Image size: Italic support and Incorrect image size when using subscript

gsomix:
Kudos to Marc Palmans!

[Image size: Italic support](https://bugs.launchpad.net/wpf-math/+bug/1046669)
> Image size is incorrectly calculated when using italic.
> I'm not sure if this patch 100% correct but it seems to work for the samples I tested.

[incorrect image size when using subscript](https://bugs.launchpad.net/wpf-math/+bug/1041219)
> The bitmap size is calculated incorrectly when using subscript.

## Example:

a_{b_{c_d}} before
before patch

a_{b_{c_d}} after
after patch

some \\, text \\, P before
before patch 2

some \\, text \\, P after
after patch 2

#3

### DIFF
--- a/WpfMath.Example/MainWindow.xaml.cs
+++ b/WpfMath.Example/MainWindow.xaml.cs
@@ -1,6 +1,7 @@
 ﻿using Microsoft.Win32;
 using System;
 using System.Collections.Generic;
+using System.IO;
 using System.Linq;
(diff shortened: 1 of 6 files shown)


ForNeVeR:
Thanks!

PR #19, opened 2017-02-13
Fixes from the old repository

ForNeVeR:
These're some commits @alexreg sent me by email. They were initially done in his own clone of the old Launchpad repository. In particular, we know that they should fix [bug #1094269](https://bugs.launchpad.net/wpf-math/+bug/1094269) and [bug #832610](https://bugs.launchpad.net/wpf-math/+bug/832610) but these commits are probably generally useful.


ForNeVeR:
Rebased the only commit with changes to `master`.

Issue #14, opened 2017-02-12
Extensible delimiters (\left and \right)


gsomix:
Full brace list we need to support:
- [x] `\left.` `\right.` (empty delimiter)
- [x] `\left(` `\right)`
- [x] `\left[` `\right]`
- [x] `\left\{` `\right\}`
- [x] `\left<` `\right>`
- [x] `\left|` `\right|`

### Plan

- [x] take a look at Java implementation — _at least JLatexMath seems to support those delimiters ~ F_
- [x] implement delimiters as a resizable characters (i.e. don't split them into top / top-middle / etc. parts yet)
- [x] add full support with splitting the characters into parts — _it suddenly turned out that we already have those_

ForNeVeR:
I think that we'll need to store brace stack somewhere when interpreting the formula, and try to keep their sizes equal.

alexreg:
This is definitely a feature that should be prioritised. I'm not sure if the Java library I ported from handled this at all (actually there were two Java libraries; one for plain TeX and one for LaTeX), but it's worth a check. If not, having a look at the LaTeX source code may be informative.

ForNeVeR:
Aha, it seems like [jlatexmath](https://github.com/opencollab/jlatexmath) have support for `\left` and `\right`.

ForNeVeR:
Just dumping what I've found regarding this task so far.

It seems that we have the atom definition for the brace-enclosed formula; it's called [`FencedAtom`](https://github.com/ForNeVeR/wpf-math/blob/4f9ad0f0ba365bfe4c5d76446ae020bf23cb0ddc/WpfMath/FencedAtom.cs).

It looks like that `FencedAtom` isn't used in the current code, but it could be exactly what we need. I plan to add the support for `\left` and `\right` commands into [`TexFormulaParser.ProcessCommand` method](https://github.com/ForNeVeR/wpf-math/blob/f1c61cf32952146b8b9418bc6e3783f71c1a8d61/WpfMath/TexFormulaParser.cs#L242). It seems like that won't be very hard: just parse the formula part before the corresponding `\right` delimiter and enclose it in `FencedAtom` (probably we'll want to keep that stack somewhere; we'll see).

I also want to add unit tests for the parser; I'll file another issue about that.

alexreg:
Good work on this. I haven't looked at the implementation yet, but I'm slightly concerned that most of the brackets are implemented except for `\left|`. Did you not do this in a generic way?

[...]

PR #20, opened 2017-02-13
Minor fixes

ForNeVeR:
These're changes remaining after rebasing #19. I've checked them out locally, and I don't think I like the new big parens. Please compare yourself:
image (old) vs image (new)

I think we should reject the patch and go with #14 instead. @gsomix, what do you think?


ForNeVeR:
Alright, let's focus on #14 then.

PR #22, opened 2017-02-15
Consistent naming

ForNeVeR:
That "issue" bugged me for a while, so I've finally decided to fix it.

We'll now use the project official (display) name "WPF-Math" everywhere. And the project technical name will be `WpfMath` (it's the name of the solution file, namespaces, NuGet package).

Also I've removed the `\r` from the solution file; I believe that's the right thing to do.


ForNeVeR:
Thanks for the review,

PR #27, opened 2017-02-17
Add parser tests

### LINKED ISSUE #21: Add parser tests
My general experience about parsing dictates one simple rule: **every parser should have a set of unit tests.** Parsers are usually very complex beasts, but it's not that hard to test them: you just pass in some text sequences and compare parser output with stuff you expect.

I think that we'll add unit tests (and will set up CI to execute them) in #14.

ForNeVeR:
So, I've decided to add some rudimentary tests and tune up the infrastructure to further improve them as the new requirements (e.g. #14) arrive.

Closes #21.


ForNeVeR:
Thanks.

PR #31, opened 2017-02-21
Moved all source and project—related files into `src/` directory.

alexreg:
I think this cleans up the root directory, and adopts a more standard project directory structure. Admittedly this isn't a Unix-based project, but I think it's a nice convention anyway. FYI everything gets built into `bin/` in root. If you don't like it, no worries!

### Remaining things to do _(these're notes for me ~ F.)_

- [x] Check `bin` placement more closely — _seems to be okay_
- [x] ~~Fix unit testing script~~ — _not required, it already works_
- [x] Fix `nuspec`
- [x] ~~Fix publishing script~~ — _not required_


ForNeVeR:
First of all, I approve the idea overall. It seems like a nice project layout: put main build files and general documentation (readme, license) into the root folder, sources to `src/`, docs to `docs/` etc.

Although I have some small nitpicks about the implementation. I'll mention them in a review section.

Also I can see that you haven't changed the test and publication scripts; that's okay, you don't need to address these. I'll take care of them myself in this branch after we settle other issues.

LICENSE
```diff
@@ -0,0 +1 @@
+LICENSE.md
```
ForNeVeR: Are you sure we need these files at all?

REAMDE
```diff
@@ -0,0 +1 @@
+README.md
```
ForNeVeR: 1) you have a typo in this file's name: `REAMDE` ;)
2) the same as `LICENSE`: I don't think we need these

alexreg: LICENSE is a symlink to LICENSE.md, and README a symlink to README.md. I quite like it that way, but if you'd rather not, no worries...

alexreg:
Regarding the placement of the sln file, that makes sense to me; let's move it to the root directory.

If you could sort out the test and publication scripts (and modify the AppVeyor spec too, if needed), then that would be great, since I'm not very familiar with how they work.

REAMDE
ForNeVeR: Ah, that's a symlink. I've never seen one in GitHub diff; that's new for me.

Honestly I'd prefer to not bother the project with symlinks. You know, we still have a plenty of troubles with them on Windows even in 2017 😞 (and Windows is still a major development and environmental platform for WPF-Math, even if we're as willing to go cross-platform as we are).

I think I'll delete these symlinks if that's okay for you, and fix the testing/publishing scripts.

alexreg: That's alright. I just add them out of pedantic convention, I suppose... although I suspect there was once a better reason that I got into such a habit! And you're wright: symlinks are a pain on Windows still.

Anyway, I'll leave the rest to you then?

ForNeVeR: Yeah, I'll finish it now.

ForNeVeR:
Thanks for the pull request.

Issue #38, opened 2017-02-23
Cannot change text style


gsomix:
We have \mathrm, \mathit, \mathcal command, but it doesn't affect on rendering.

We should fix it first, before implementing #32 as math functions (sin, cos, log) have \mathrm style.

ForNeVeR:
Fixed by #39 I believe.

PR #39, opened 2017-02-23
TexFormulaParser: Propagate text style through parsing

gsomix:
Fix for #38 

mathrm

### DIFF
--- a/WpfMath.Tests/ParserTests.fs
+++ b/WpfMath.Tests/ParserTests.fs
@@ -15,7 +15,12 @@ type ParserTests() =
             .ExposeInternalsOf<FencedAtom>()
             .Assert()
 
+    let rmStyle = "mathrm"
+    let itStyle = "mathit"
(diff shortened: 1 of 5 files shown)


WpfMath.Tests/ParserTests.fs
```diff
@@ -51,4 +56,37 @@ type ParserTests() =
...
+
+    
+    [<Fact>]
+    let ``\mathrm should be parsed properly`` () =
+        assertParseResult
+        <| @"\mathcal{sin}"
+        <| (formula <| row [styledChar('s', calStyle); styledChar('i', calStyle); styledChar('n', calStyle)])
+
+    [<Fact>]
+    let ``\mathrm{} should be parsed properly`` () =
```
ForNeVeR: I think that this test name should be "\mathrm{} should not be parsed"

ForNeVeR:
Thanks!

PR #41, opened 2017-02-23
Add \lim command

gsomix:
Just trying \mathrm for new math commands...

lim


ForNeVeR:
Wow, very nice. Thanks!

PR #48, opened 2017-02-24
Operators-symbols should be parsed as BigOp

### LINKED ISSUE #24: Integral sign displays incorrectly
Affects only for indefitite integral formula, `$$ \int_{}^{} $$` works fine. 

int error

gsomix:
Fix #24 

int


ForNeVeR:
Thanks!

Issue #50, opened 2017-02-24
Improve blurred output


torfranz:
Hi, 
Currently formulas are rendered pretty blurred.

Toying with the v0.3 version now it seems I'm unable to influence the rendering of the formula with any of the usual options, RenderOptions, TextOptions, UseLayoutRounding. I tried basically all combinations but to no avail. Can you give any hint on how to reduce the blur?

Btw, thx for the lightning speed in releasing v0.3 👍

ForNeVeR:
Alright, it wasn't completely fixed. See comment from @torfranz:

>![image](https://cloud.githubusercontent.com/assets/13871318/23846930/3bd3d942-07d0-11e7-9eb6-6f2d5ec784af.png)
>
> Please note the blurred horizontal lines on the equal sign.

ForNeVeR:
That's not a font issue, it's more of WPF aliasing / layout rendering / pixel pinning issue. You can't just "fix" it by changing a font.

Also, for the record, the right one looks more blurry to me (check the radical sign: its' top part definitely has more pixels in height than it should; the top pixel line is shadowed).

B3zaleel:
Have you tried overriding the OnRender method of the formula control and changing the VisualTextRenderingMode of the base to TextRenderingMode.Cleartype or GrayScale.

B3zaleel:
You could make the following changes in the "WpfMath/Rendering/WpfElementRenderer.cs" file.
```csharp
public void RenderGlyphRun(Func<double, GlyphRun> scaledGlyphFactory, double x, double y, Brush foreground)
        {
            var glyphRun = scaledGlyphFactory(_scale);
            Pen glyphPen = new Pen(foreground, 1);
            _drawingContext.DrawGeometry(foreground, glyphPen, glyphRun.BuildGeometry());

            //_drawingContext.DrawGlyphRun(foreground, glyphRun);
        }
```
It has the tendency of being too bold so you'll have to make a lot of small changes to it(the Pen).

PR #52, opened 2017-02-25
[Discussion in progress] Round all the calculations to whole pixels

ForNeVeR:
While investigating #50 I've found that we have small rounding issues everywhere. To fix them, I had to introduce rounding in every `Draw` method.

**Obviously we could add a smarter solution, but let's discuss that in principle at first.**

The outcome is that the formulae are not so blurry. Check out these screenshots:
Before: image

After: image

Unfortunately, now we have small troubles _(on some of the scale levels)_ when connecting part of square root tail with the rest of the radical line. Honestly said, earlier we had _small_ issues there, but now we have huge noticeable gap. The following examples were taken with `Scale = 21.0`:

Before:  image

After: image

That last issue is cleanly a regression, so I'm not sure if we want to merge this PR as-is. Let's discuss! @torfranz, I welcome you to take a look, too.

### DIFF
--- a/src/WpfMath/CharBox.cs
+++ b/src/WpfMath/CharBox.cs
@@ -39,6 +39,9 @@ namespace WpfMath
 
         public override void Draw(DrawingContext drawingContext, double scale, double x, double y)
         {
+            x = TexRenderer.RoundToWholePixels(x, scale);
+            y = TexRenderer.RoundToWholePixels(y, scale);
+
             base.Draw(drawingContext, scale, x, y);
 
             GlyphRun glyphRun = GetGlyphRun(scale, x, y);
--- a/src/WpfMath/Controls/FormulaControl.xaml.cs
+++ b/src/WpfMath/Controls/FormulaControl.xaml.cs
@@ -12,7 +12,7 @@ namespace WpfMath.Controls
     public partial class FormulaControl : UserControl
     {
         private static TexFormulaParser formulaParser = new TexFormulaParser();
-        private TexFormula texFormula; 
+        private TexFormula texFormula;
 
         public string Formula
         {
@@ -45,7 +45,7 @@ namespace WpfMath.Controls
         }
 
         public static readonly DependencyProperty FormulaProperty = DependencyProperty.Register(
-            "Formula", typeof(string), typeof(FormulaControl), 
+            "Formula", typeof(string), typeof(FormulaControl),
             new PropertyMetadata("", OnFormulaChanged, CoerceFormula));
 
         public static readonly DependencyProperty ScaleProperty = DependencyProperty.Register(
@@ -84,7 +84,7 @@ namespace WpfMath.Controls
 
             using (var drawingContext = visual.RenderOpen())
             {
-                renderer.Render(drawingContext, 0, 0);
+                renderer.Render(drawingContext);
             }
             formulaContainerElement.Visual = visual;
         }
--- a/src/WpfMath/GlueBox.cs
+++ b/src/WpfMath/GlueBox.cs
@@ -31,6 +31,8 @@ namespace WpfMath
 
         public override void Draw(DrawingContext drawingContext, double scale, double x, double y)
         {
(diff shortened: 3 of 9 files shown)


alexreg:
Hmm. I can see there was already a minute issue, like you say. Did you figure out the cause of this (roughly speaking, even)?

alexreg:
I have no idea how to solve this problem elegantly right now (perhaps the hour is to blame!)... but I'm not a great fan of the current solution, to be honest. Did you consider any other approaches?

ForNeVeR:
@alexreg thanks, that was a good question. I've struggled a bit around SVG rendering code, and found that the SVG image rendered by our code had no flaws. So, I've tried to replace the code in `VerticalBox.Draw` with a call to `RenderGeometry`:

```csharp
    internal class VerticalBox : Box
    {
        // ...
        public override void Draw(DrawingContext drawingContext, double scale, double x, double y)
        {
            base.Draw(drawingContext, scale, x, y);

            var g = new GeometryGroup();
            RenderGeometry(g, scale, x, y);
            drawingContext.DrawGeometry(Brushes.Black, new Pen(), g);
        }
        // ...
    }
```

And here's what I got:
image

The radical line is now absolutely straight in any scale, but
1) the vertical line in the radical sign have noticeable gaps between its composable parts
2) the font antialiasing was completely disabled, because WPF now renders actual geometry instead of font characters _(although I'm not sure that's a bad thing; the result without ClearType actually looks a bit clearer to me)_

I think that's a viable solution, but we need more work on it:

1) fix these gaps on the vertical line
2) try to scope the change only to the radical sign and surrounding lines (everything else should stay as-is)

We need also think more about switching the whole render to `DrawGeometry` instead of `DrawGlyphRun`. Probably that will be the best solution.

alexreg:
Ahh, those damned gaps in the line — I actually remember them from my testing 6 years ago!

I think that turning off ClearType would perhaps be a good idea, since TeX is not designer to work with it.

alexreg:
Just had a little play around, and no idea what's causing those gaps! If I couldn't solve it back then... do have a go, and let me know how you get on though.

torfranz:
Hi,
I tried the 2 branches 50-blur and 50-blur-alternate against the master and for my particular case 50-blur was the best option so far. My formulas are rather simple compared to the example formula so I can't really tell what will be best going forward. Maybe some **RenderPath** property as long as there is no best universal solution? Thanks for all your effort so far, I see this will probably not be easily solved.

ForNeVeR:
Idea from #66: compare our fonts with the ones from LaTeXMath — probably the guys there have already solved the issue.

alexreg:
Yep. If using their fonts doesn't work, and there are no obvious tricks in their code, perhaps we can even point them to our issue, and see if they recognise the problem from their past work.

gsomix:
Lastest `50-blur-alternate` branch.

big

small

ForNeVeR:
As we ascertained with @gsomix, now that branch looks nice, but still blurred at some of the scale levels. Probably we'll need to merge it with (at least some of) the changes in `bugfix/50-blur`.

alexreg:
Very good. Yeah, I'll let you test that out. If we could be clear about what changes we're making to reduce blurriness (and make the minimal number possible), then that would be great. Would be happy to review a PR then.

ForNeVeR:
Yep. We're slowly making progress in this issue, it's not yet ready for review and merging.

ForNeVeR:
I'm closing the PR and deleting the branches `bugfix/50-blur` and `bugfix/50-blur-alternate`, because issue #50 was solved by #71.

alexreg:
Small issue, can probably be committed directly: change default formula in example to:

```latex
L = \int_a^b \sqrt[4]{ \left| \sum_{i,j=1}^ng_{ij}\left(\gamma(t)\right) \left[\frac{d}{dt}x^i\circ\gamma(t) \right] \left{\frac{d}{dt}x^j\circ\gamma(t) \right} \right|}dt
```

ForNeVeR:
@alexreg done that!

PR #53, opened 2017-02-25
Fix crash while parsing empty groups

gsomix:
Example: "{}"
Expected: nothing to render
Actual: crash with NRE


ForNeVeR:
Thanks for the fix!

PR #54, opened 2017-02-26
Remove delimiter type constraint

ForNeVeR:
After report by @alexreg in #14 that `|` doesn't work I decided to review the delimiter typing. I've found that things like `\left)` are actually supposed to work, so I removed the constraint on the delimiter type from the closing brace code.

That way, now things like `\left| \frac{2}{2} \right|` work, and even `\left) \frac{2}{2} \right|`.

image


ForNeVeR:
Thanks for the review.

alexreg:
Yep, I think this is the right way to go about things.

Issue #55, opened 2017-02-26
Multiple integrals


gsomix:
We need to implement multiple integrals commands.

https://www.sharelatex.com/learn/Integrals,_sums_and_limits#/Multiple_integrals

B3zaleel:
I think a way to do that would be to modify the defaultTexFont and it's dependencies.
The current fonts when I checked, don't have some character glyphs, but Asana math, and a few other fonts do. I developed a little tool for previewing the glyphs in a character. Try to get the Asana math font and put 8650 and 8800 at the start and end Textbox so and update, to see the characters.

alexreg:
@B3zaleel Thanks again for ideas and contributions. You seem very keen to take this project forward, which I'm grateful for. Unfortunately I (the original author of this project) have very little time to maintain this project these days, and rarely use .NET. @ForNeVeR is the current maintainer, but he seems rather preoccupied too. Let's see if he shows up over the next week or so, but I encourage you to maintain your own branch for now, and make comments/PRs here.

ForNeVeR:
@B3zaleel thanks for your ideas. I fear it won't be as easy as just replacing the font we're using with Asana Math (although it's a good suggestion on its own: I'd like to eventually support user-supplied fonts as well).

For WPF-Math to properly process the font, we need to extract some metrics data (check `DefaultTexFont.xml`), and for now we're not sure how to get some of the metrics from the font file in an automated way :(

B3zaleel:
I managed to get a separate version to work with Asana Math but the main problem is with the big operators and radicals. The current algorithm favors big operators with more depth than height. The memory usage is massive as well.

B3zaleel:
Check the font metrics in the "Data" directory of PR #156's patch. The file is "AsanaMath....wmpkg", it can open with 7zip and currently has over 2000 symbols.

PR #56, opened 2017-02-26
[WIP] Multiple integrals

gsomix:
Ok, please, take look at implementation of multiple integrals.

iint
idotsint

Spacing between integral symbols is accurate, but subscript looks poor.

iint_v

It seems there are specific rules for "one-symbol-operator-subscripts", I'll investigate it later.

#55

### DIFF
--- a/src/WpfMath/Data/PredefinedTexFormulas.xml
+++ b/src/WpfMath/Data/PredefinedTexFormulas.xml
@@ -244,6 +244,44 @@
     <Return name="f" />
   </Formula>
 
+  <!-- Multiple Integrals -->
+  
+  <Formula name="iint" enabled="true">
+    <CreateFormula name="f" />
+    <MethodInvocation name="AddOperator" formula="f">
+      <Argument type="string" value="{\int}\negmedspace{\int}" />
+      <Argument type="bool" value="true" />
+    </MethodInvocation>
+    <Return name="f" />
+  </Formula>  
+  
+  <Formula name="iiint" enabled="true">
+    <CreateFormula name="f" />
+    <MethodInvocation name="AddOperator" formula="f">
+      <Argument type="string" value="{\iint}\negmedspace{\int}" />
+      <Argument type="bool" value="true" />
+    </MethodInvocation>
+    <Return name="f" />
+  </Formula>  
+
+  <Formula name="iiiint" enabled="true">
+    <CreateFormula name="f" />
+    <MethodInvocation name="AddOperator" formula="f">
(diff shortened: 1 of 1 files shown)


ForNeVeR:
_I'm not sure_, but it seems to me that the integral signs should be a bit closer to each other.

alexreg:
The subscript doesn’t seem correctly positioned (horizontally) for multiple integrals.

B3zaleel:
If you convert the unicodes 8747 to 8755 to a string using Asana Math.otf, you would see what the integrals would look like.
Given the variety of symbols Asana Math contains, it might be worth looking at.

B3zaleel:
Do we have plans on supporting different packages(math fonts, symbols, formula settings, predefined formulas and colors)? because I have a local branch that does that(although I've only done a package for Asana Math and it has a variety of integrals like sqint,etc).

Happypig375:
Why would you implement using your own kerning instead of mapping the multi-integtals' codepoints to their glyphs?

ForNeVeR:
@B3zaleel yes we have discussed the idea of preparing multiple "packages" that could contain WPF-Math XML files for different fonts, formula subsets and maybe LaTeX "flavors". Although I'm concerned more about organizational stuff: how we would redistribute that, how the user will plug in these packages at runtime etc.

Overall I think that's a good idea and I'd like to approach this.

ForNeVeR:
@Happypig375 sorry but I'm not sure who and what are you asking. Who does "implement their own kerning"? Does it relate to this PR, some of @B3zaleel words, or in general to WPF-Math?

B3zaleel:
We could put those files in a zipped file, we can then initialize the parser with the file or use a default one. The files would then be extracted in the documents special folder, kept hidden and protected and then we can use it.

Happypig375:
@ForNeVeR My comment is essentially a code review of this PR.

About resource packages, I worry about conflicting packages. They are hard to debug.

ForNeVeR:
@Happypig375 well about multi-integrals: we'd need to check how it works in other LaTeX engines. What if I have more than three integral signs, for example?

Happypig375:
⨌	U+2A0C Quadruple integral :wink:

ForNeVeR:
I'm not sure that our current engine has support for that kind of character merge / ligature support. We have _some basic_ kind of ligature support, though I never had a chance to check that.

What you say make sense to me though.

gsomix:
OK, guys, I should have done this a long time ago. :)

PR #58, opened 2017-02-26
Support for empty delimiters

### LINKED ISSUE #14: Extensible delimiters (\left and \right)

ForNeVeR:
That's the last piece of #14. It allows to do things like this:

image

~~We already support rendering of `FencedAtom` with either `LeftDelimiter` or `RightDelimiter` being defined as `null`, so I've decided that we should construct it so when any delimiter is omitted.~~

~~Although our parser isn't `null`-tolerant at all, and I don't think it should be. So, parser generates symbols with a special names, and those get replaced when constructing the `FencedAtom`.~~

I've used another solution without abusing `null` values.

Closes #14.

### DIFF
--- a/src/WpfMath.Tests/ParserTests.fs
+++ b/src/WpfMath.Tests/ParserTests.fs
@@ -43,9 +43,10 @@ type ParserTests() =
     [<Theory>]
     [<InlineData(".", ")", true, false)>]
     [<InlineData("(", ".", false, true)>]
-    let ``Empty delimiters should work`` (left : string, right : string, isLeftNull : bool, isRightNull : bool) =
-        let leftBrace = if isLeftNull then null else (openBrace "lbrack")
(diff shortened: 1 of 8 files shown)


gsomix:
@ForNeVeR, also "\left. x \right y".

ForNeVeR:
@gsomix that was fixed.

PR #68, opened 2017-03-04
#66: add information about Knuth license

### LINKED ISSUE #66: Clarify font licensing
We're currently providing [some TTF files](https://github.com/ForNeVeR/wpf-math/tree/9236db231ec2528f4bc947697cb7a4c0f812d623/src/WpfMath/Fonts), but I can't find their license (it'd be unusual to find that the fonts are licensed by GPL, too). Some of them looks like [Computer Modern](http://www.ctan.org/texarchive/fonts/cm) (which is distributed under something marked as "Knuth license"). We need to find the source and properly attribute these files.

ForNeVeR:
That's the last missing piece of information about the font licensing.

Closes #66.


ForNeVeR:
Thanks for help!

PR #69, opened 2017-03-05
[WIP] Gotta draw 'em all on Geometry!

gsomix:
The idea is to render formula in vector geometry. It allows us to join complex symbols (e.g. big delimiters, sqrt symbol) and to place them onto DrawingContext taking into account WPF DIP. Hopefully such method will reduce blur.

Check list!

- [x] Render all -boxes on Geometry
- [ ] Join intersecting geometries
- [ ] Add rounding
- [ ] Fix \color
- [ ] Fix SVG export

### DIFF
--- a/src/WpfMath/Box.cs
+++ b/src/WpfMath/Box.cs
@@ -89,16 +89,6 @@ namespace WpfMath
             set;
         }
 
-        public virtual void Draw(DrawingContext drawingContext, double scale, double x, double y)
-        {
-            if (this.Background != null)
-            {
-                // Fill background of box with color.
-                drawingContext.DrawRectangle(this.Background, null, new Rect(x * scale, (y - Height) * scale,
-                    (this.Width + this.Italic) * scale, (this.Height + this.Depth) * scale));
-            }
-        }
-
         public virtual void RenderGeometry(GeometryGroup geometry, double scale, double x, double y)
         {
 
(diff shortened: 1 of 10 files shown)


alexreg:
We'll want to move the SVG "export" into a rendering library entirely separate from WPF soon, right?

ForNeVeR:
@alexreg current "SVG exporter" is entirely a user-level feature (i.e. WPF-Math users should implement it on theirs side, although we provide an example); it relies on WPF `GeometryGroup` render. Summary:

- we haven't any SVG-specific code in WPF-Math, but we're just providing `GeometryGroup` render
- SVG and PNG support currently can't be decoupled from WPF because it relies on WPF geometry objects
- rendering into `GeometryGroup` and drawing into `DrawingContext` are two distinct operations, and users should choose between them depends on whether they want to export image (to SVG or whatever) or draw it on screen
- this PR unifies the operations and leaves only the `GeometryGroup` mode, because we can render it into `DrawingContext` without any special functions (the resulting image looks slightly different, but our current consensus is that looks better)

The purpose of this is not only refactor the library and deduplicate the code, but also to provide a possibility to optimize the image drawn. We will conduct some experiments with snapping the positions of resulting `Path` graphical nodes into screen pixels of the target device (like #52, but in regards to `Geometry`). Hopefully, it will provide less blurry results than the current implementation.

ForNeVeR:
After thorough discussion, we've decided to merge f82f294e789bf05edf6a47b4a9d8f58f03af031f and drop everything else from this PR for now.

PR #71, opened 2017-03-11
Use guidelines for WPF render

ForNeVeR:
Fixes #50.

That's my another try to fix that long-standing issue. That time I've finally read the documentation (thanks @Atomosk for pointing me to the tutorial page [Draw lines excactly on physical device pixels](https://wpftutorial.net/DrawOnPhysicalDevicePixels.html)).

That time, we have no obvious regressions for any of the scale settings, so I think that's definitely a way to go. Check the comparison (_clickable_):

image

(The fonts are notably clearer now, and even without turning off the font antialiasing settings!)

These changes supersedes the changes in #52 and in the branches `bugfix/50-blur` and `bugfix/50-blur-alternate`; gonna to delete them after the merge.

PR #72, opened 2017-03-12
TexRenderer: abstract guidelines from drawing code

ForNeVeR:
That's a follow-up for #71. I've decided to abstract the guidelines code from the `Draw` method, and implement the [Template method pattern](https://en.wikipedia.org/wiki/Template_method_pattern) (with `DrawWithGuidelines` being the template method and `Draw` being the primitive) for the following purposes:

- keep `Draw` clean from the guidelines code
- regain principial compatibility with #69 
- remove the need to call `base.Draw` in `Box`-derived classes

Now every box uses `DrawBackground` method and every box supports guidelines (my previous implementation worked only for `HorizontalBox` and `VerticalBox`).


ForNeVeR:
Yep, sure. Thanks for the review!

Issue #62, opened 2017-02-26
Delimiters should support scripts


gsomix:
`\left.\frac{x^3}{3}\right|_0^1`

![del_scripts](https://cloud.githubusercontent.com/assets/832449/23340671/d229a76a-fc53-11e6-904c-81a77033d410.PNG)

ForNeVeR:
Another example from #73:

```tex
d_{f,t}(u,v) = \sum_{x, y}\left(f(x, y) - t(x - u, y - v)\right)^2
```

ForNeVeR:
Fixed by #74,

PR #74, opened 2017-03-17
Add support for scripts to delimiters

gsomix:
Fix for #62 

ex1

ex2

cc @illiteratePhD


ForNeVeR:
@alexreg I think that'll contradict the default LaTeX behavior. While I can clearly see why it's useful, I'm not sure we need to introduce such behavior. Although we can make it optional.

@gsomix thanks for the contribution, that's very nice.

PR #75, opened 2017-03-18
Project: release 0.3.1

ForNeVeR:
I suggest we publish a new version today. Any objections?


ForNeVeR:
Thanks for the review!

PR #79, opened 2017-04-08
SVG: Added support for curves and removed flattening of the geometry

nocturnals:
Second attempt.. *grin*
Feel free to criticize & correct.

I'm not a fan of the if(.. is polywhatever) and then .. as polywhatever. constructs (casting twice)
Not sure how to fix it while maintaining readability.

I was also thinking about ditching the entire RenderGeometry thing. It creates more problems than it solves. For example I wanted to include colors in the svg but the geometry only provides path data not colors.

edit: I forgot: currently the geometry returned by BuildGeometry auto converts to cubic bezier curves while truetype fonts has quadratic curves. I've added the case for quadratic curves but it's untested. Should you want to use a more general portable non-wpf (freetype, for example) method of reading/rendering fonts, it's already in there and should work.

### DIFF
--- a/src/WpfMath/CharBox.cs
+++ b/src/WpfMath/CharBox.cs
@@ -46,8 +46,7 @@ namespace WpfMath
             GlyphRun glyphRun = GetGlyphRun(scale, x, y);
 
             GeometryGroup geoGroup = glyphRun.BuildGeometry() as GeometryGroup;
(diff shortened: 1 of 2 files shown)


ForNeVeR:
Old part of the sigma sign (first) vs. the newer part (second):

image image

@nocturnals, that's excellent, thank you.

PR #80, opened 2017-04-08
force streamwriter to flush buffers

nocturnals:
When I saved an svg file not everything was written on my computer.


ForNeVeR:
Thanks for the contribution!

PR #86, opened 2017-06-01
UTF-8 support in parser and render

### LINKED ISSUE #82: Support UTF8
Can you support UTF8 symbols? I can't add symbols from other languages than English!

ForNeVeR:
This is an attempt to use system fonts in our render.

@nocturnals, you could take a look at my `ITeXFont` implementation. I can see that you're trying something similar for OpenType; probably you'll find my approach to `SystemFont` useful (I've implemented the necessary minimal amount of methods for now).

Closes #82.

Here's what it looks like at the moment (yep, it's imperfect, but I suggest to merge it and wait for further reports and/or maybe help from the contributors):

image


alexreg:
Are you trying to move things over to the TeX.NET parser too, which I know you integrated into this repo a little while back?

ForNeVeR:
@alexreg, nope, it's not yet time of that parser. The main problem of the current task is the rendering and not the parsing. That means that current task won't get any benefits from new TeX parser, and that's why I'm not integrating it right now.

Integration of TeX.NET parser is in my backlog (more precisely, I'm going to inspect its' code closer and decide the benefits of that integration), but currently it resides in its' own repository: https://github.com/ForNeVeR/mathtexdotnet

alexreg:
Fair enough. I just thought while you were touching the parser code, you might want to rewrite it anyway, based on MathTeX.NET. But no worries; whichever way you think is best!

nocturnals:
> I can see that you're trying something similar for OpenType; probably you'll find my approach to SystemFont useful

Not really, there were a lot of issues to implement opentype. If there was only one thing that would require a major rewrite, I would have suggested it and see what the reactions were. So I decided in the mean time to put it on hold.

These are the issues I can remember on the top of my head:
## Layout:
Opentype math was constructed using tex as a reference but they expanded it. This means that you have more parameters to use to layout. There are two methods that current tex engines use: either expand their internal parameters or map the opentype params to tex (luatex uses one method while xetex uses the other).

## Font information:
### GlyphTypeface
Not all info needed is contained in GlyphTypeface. It's actually a bit worse than that. Some info is there but marked private. In any case, I think in the long run we'll need to replace it by something similar like freetype.

### Freetype
Even in the case should we switch to freetype, we're still not there yet. In the test program I made, I parsed parts of the font file myself. Freetype doesn't have any support to read the math table of opentype. So we are write our own parser or implement it in freetype.

It's serendipitous that @alexreg mentioned a new parser. When I first started this was one of first things I would implement. I haven't looked at the code but from his description it sounds like there's greater divide between the atoms and other parts of the code. This would make it easier to implement things like: layout using tex rules or layout using opentype rules. I could then implement opentype without impacting the tex code.

I was using the doc below as my main source of info on opentype math:
[OpenType Math Illuminated](https://www.ntg.nl/maps/38/03.pdf)

ForNeVeR:
I'd like to merge that in its' current state. @gsomix, review please!

PR #89, opened 2017-09-10
Fix accent alignment

### LINKED ISSUE #88: Bar alignment
In my statistical application I need to show a double bar over one character like in image

doublebarx
 
I try  the syntax \bar{\bar{X}} but the bars are not aligned

alignbar

ForNeVeR:
Closes #88.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,4 @@
+/.idea/
 /.vs/
 /packages/
 
--- a/src/WpfMath.Tests/BoxTests.fs
+++ b/src/WpfMath.Tests/BoxTests.fs
@@ -1,0 +1,23 @@
+namespace WpfMath.Tests
+
+open DeepEqual.Syntax
+open Xunit
+
+open WpfMath
+open WpfMath.Exceptions
+open WpfMath.Tests.Utils
+
+type BoxTests() =
+    [<Fact>]
+    let ``AccentedAtom should have a skew according to the char`` () =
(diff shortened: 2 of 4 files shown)


.gitignore
```diff
@@ -1,3 +1,4 @@
+/.idea/
```
alexreg: This should be put in a local (non-committed) gitignore, I think.

ForNeVeR: And I would argue: at the moment, Rider is one of the major development tools I personally use for the project. If we put Visual Studio stuff there, we should put Rider stuff as well. That'll be just convenient for the people working with the project: they won't need to set up anything to ignore any stuff their IDEs are creating.

Before anybody ask: I'd be happy to add MonoDevelop (or Xamarin Studio, or something else) stuff in our .gitignore, too, if somebody uses that IDE.

alexreg:
Okay. I generally think that unless it is the official software for dealing with files in the repository, it should not belong in the .gitignore, but whatever you like… no big deal.

Go ahead and merge, otherwise! :-)

ForNeVeR:
No problem. It's always good to hear alternate opinion on the matter.

ForNeVeR:
@gsomix, next time please wait for me to merge the PR. You've done **nothing wrong**, and everything is **okay**, but I usually prefer to implement more complex merge strategy than just, well, _merge_.

ForNeVeR:
Ah, and thanks for the review, guys!

PR #93, opened 2017-11-27
Extract Renderer layer

### LINKED ISSUE #92: Extract Renderer code to a single class/interface
Currently there's a `Box.Draw` method that's implemented in a relative small number of classes.

For further extensibility, I'd like these methods to be extracted to the `IRenderer` interface and `WpfRenderer` class using the Visitor pattern (or something alike).

In future, we'll likely extract `Box.RenderGeometry` (that has the same form as the `Draw` method) to another `IRenderer` implementation, `GeometryRenderer`. That will also open a way for alternate implementations, such as `DirectXRenderer` and `AvaloniaRenderer`.

ForNeVeR:
Closes #92.

**Things remaining to do**:

- [x] extract `RenderGeometry` method to `GeometryRenderer` class
- [x] add some tests! (with maybe a `TestRenderer`?)

**Open questions**:

- [x] I'd like to uncouple `IElementRenderer` from WPF's `Transform` type. Maybe replace it with our own abstraction that will be converted into WPF `Transform` or just `Translate` / `Rotate` method calls on the `GeometryRender`?
- [x] ~~I don't like the first argument of `RenderGlyphRun` (because it's a lambda). Maybe we could somehow replace that with `GlyphRun` and scale it in the renderer code?~~
  _Added a TODO comment for that, too hard to implement at the moment._

**Impact**:

As I said in #92, this opens a supported way to implement alternate renderders for DirectX, Avalonia and whatnot. There's still more work to do with `GlyphRun` replacement, but I think we'll do that later.

### DIFF
--- a/src/WpfMath/CharBox.cs
+++ b/src/WpfMath/CharBox.cs
@@ -31,15 +31,6 @@ namespace WpfMath
                 new ushort[] { glyphIndex }, new Point(x * scale, y * scale),
                 new double[] { typeface.AdvanceWidths[glyphIndex] }, null, null, null, null, null, null);
             return glyphRun;
-
-        }
-
-        public override void RenderGeometry(GeometryGroup geometry, double scale, double x, double y)
-        {
-            GlyphRun glyphRun = GetGlyphRun(scale, x, y);
-
-            GeometryGroup geoGroup = glyphRun.BuildGeometry() as GeometryGroup;
-            geometry.Children.Add(geoGroup);
         }
 
         public override void RenderTo(IElementRenderer renderer, double x, double y)
--- a/src/WpfMath/CharBox.cs
+++ b/src/WpfMath/CharBox.cs
@@ -34,14 +34,6 @@ namespace WpfMath
 
         }
 
-        public override void Draw(DrawingContext drawingContext, double scale, double x, double y)
-        {
-            GlyphRun glyphRun = GetGlyphRun(scale, x, y);
-
-            // Draw character at given position.
-            drawingContext.DrawGlyphRun(this.Foreground ?? Brushes.Black, glyphRun);
-        }
-
         public override void RenderGeometry(GeometryGroup geometry, double scale, double x, double y)
         {
             GlyphRun glyphRun = GetGlyphRun(scale, x, y);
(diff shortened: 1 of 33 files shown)


src/WpfMath/Rendering/GeometryRenderer.cs
```diff
@@ -0,0 +1,65 @@
...
+
+namespace WpfMath.Rendering
+{
+    /// <summary>A renderer that renders the elements to a provided <see cref="GeometryGroup"/> instance.</summary>
+    public class GeometryRenderer : IElementRenderer
+    {
+        private readonly double _scale;
+        private readonly GeometryGroup _geometry;
+
+        public GeometryRenderer(double scale, GeometryGroup geometry)
```
ForNeVeR: - [ ] pass `scale` as a second element everywhere

src/WpfMath/Rendering/GeometryRenderer.cs
```diff
@@ -0,0 +1,65 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Windows;
+using System.Windows.Media;
+
+namespace WpfMath.Rendering
+{
+    /// <summary>A renderer that renders the elements to a provided <see cref="GeometryGroup"/> instance.</summary>
+    public class GeometryRenderer : IElementRenderer
```
ForNeVeR: - [ ] → `GeometryElementRenderer`?

src/WpfMath/CharBox.cs
```diff
@@ -22,31 +23,20 @@ public CharInfo Character
             private set;
         }
 
-        private GlyphRun GetGlyphRun(double scale, double x, double y)
+        internal GlyphRun GetGlyphRun(double scale, double x, double y)
```
gsomix: `GlyphRun` is WPF class. Please don't forget create issue to move this part to WPF renderer.

ForNeVeR: - [ ] also, this particular line (changing of `private` to `internal`) seem unnecessary?

Thanks for your review!

ForNeVeR: > this particular line (changing of `private` to `internal`) seems to be unnecessary?

Nope, it doesn't. This method is called from the test code.

alexreg:
Good job, guys!

PR #94, opened 2017-12-02
Tests: make test functions public

ForNeVeR:
It turned out I was defining test functions as `internal` (the default for in-class `let` bindings in F#).

While generally working, this approach causes some tooling issues, so I've finally decided to fix that.

_(Also there was unnoticed function name duplication: see `\mathcal should be parsed properly`.)_


ForNeVeR:
Thanks for the review.

Issue #17, opened 2017-02-12
Avalonia support


gsomix:
[Avalonia](https://github.com/AvaloniaUI/Avalonia)

## TODO
- [x] create a branch for Avalonia stuff (and we will then proceed to work on that branch using PRs with the usual branching policy, almost as with the `master`): [`feature/17-avalonia`](https://github.com/ForNeVeR/wpf-math/tree/feature/17-avalonia)
- [x] make PR #98 non-destructive (because currently it destroys stuff, and that's unwanted) and merge it into the branch
- [x] fix builds and add CI support for both modes (WPF and Avalonia)
- [x] rebase work of @ahopper from #170 on top of the branch
- [x] make sure tests work for WPF part
- [x] make `FormulaBlock` public
- [x] solve #63 in the branch
- [x] remove the code duplication (at least most of it) — that's why #63 was necessary
- [x] update the documentation
- [x] after all of that, Avalonia may land into the `master` branch
- [x] add packaging for Avalonia, but explicitly mark the packages as alpha (#192)
- [x] check issue #128: it should'nt be reproducible for WPF and for Avalonia
- [x] currently we're planning to only allow for solid colors in the portable code base; we'll have to think about supporting for non-solid colors (aka `Brush`es) for WPF- and Avalonia-dependent code
- [x] Make sure the resources (`.xml` and `.ttf`) are deduplicated properly.

ForNeVeR:
It isn't still clear what kind of support we can offer to Avalonia. Probably we could add the control like your `FormulaControl` (see #23). Anyways, it seems that it depends on #18.

gsomix:
We can use drawing infrastructure of Avalonia that looks similar to WPF's one.

https://github.com/AvaloniaUI/Avalonia/tree/master/src/Avalonia.Visuals/Media

I will look into it later.  And yes, I want to make a similar user control but for the xplat UI framework.

charlesroddie:
Happy to have discovered this project.
For cross-platform usage (which is essential for me) may I recommend a SkiaSharp backend?
This can render on WPF, iOS, Android, UWP (and others) and can be used from Xamarin.Forms.
Avalonia is still in alpha status and if it matures and gets any significant traction 

Personally this would make the project usable for me, and would open up the possibility of integrating LaTeX into other forms of graphics.

Happypig375:
Would make sense to support Xamarin Forms at the same time too.

ForNeVeR:
[...]

PR #98, opened 2017-12-26
[WIP] Port to Avalonia

ForNeVeR:
Still very work-in-progress.

**Notes from the author:**
- [ ] Custom fonts are not working (math symbols are not working)
- [ ] Positions of the formula symbols are incorrect
- [ ] Project Properties removed as interrupting build

**Notes from me:**

We definitely don't want to remove WPF code, and we want Avalonia and WPF to coexist in our project, so the users could choose the UI framework they prefer. We'll need to put WPF code back, and extract Avalonia-specific parts from the main code base.


alexreg:
A good start!

Once this is done, a rename of the project is probably in order. Something that conveys it's a generic maths rendering library... or alternatively a more innovative name, only loosely connected to its purpose.

ahopper:
Just wondered if there was any plan to take this forward.  Avalonia now supports font embedding so that should help.

alexreg:
@ahopper Not much progress I believe. @ForNeVeR doesn't have as much time these days, I think (though he should correct me if I'm wrong)... so if you want to take up this project, that would be great. I'm sure one of us could review.

ahopper:
@alexreg I've had a quick play and with the latest Avalonia and a couple of tweaks the basics seem to work
image
I'm not sure what to do with this next.  I don't have the time right now to properly reintegrate into the wpf stuff. I also noticed some mention of combining with CSharpMath which I guess might take a different approach to avalonis integration, is this likely?

alexreg:
@ahopper The unification with CSharpMath will only be at the level of atoms, at least initially, I believe. Anyway, good work there. If you can eventually create two different rendering/front-end projects, one for WPF and one for Avalonia, that would be ideal.

ahopper:
I've put a first cut of combining the wpf and Avalonia code here https://github.com/ahopper/wpf-math/tree/avalonia-wpf-combined.
I have left the wpf code untouched and put anything that could be shared without modifying in a shared project. It is based on the current wpfmath.
The Avalonia version is more like the wpf version than this pr.
Obviously very much more code can be shared with some tweaking of the wpf code.
Before I go any further does this approach seem sensible?

ForNeVeR:
@ahopper yes your approach looks pretty great. We definitely will need more unification between WPF and Avalonia code bases, but that's very good first steps.

I believe we could merge some of your code whenever you think it's in a good state, and maybe publish a separate experimental package with Avalonia stuff. After that we'll be able gradually unify the code (e.g. extract the things we can't share between WPF and Avalonia, and merge/deduplicate everything else). That will break a lot of stuff in that _experimental_ Avalonia package in the new versions, but that'll be okay I guess.

ForNeVeR:
@victm thank you much for your contribution. It has been moved forward with #175 (and I've done my best to preserve the credits to you in the updated commit). Check #17 if you want to follow the Avalonia development in WPF-Math.

PR #102, opened 2018-01-21
Fixed a crash when rendering unsupported characters like "Å".

sskodje:
This fixes a crash when inputting characters or symbols not supported by the font. The issue was discovered using norwegian letters "æ, ø or å".

### DIFF
--- a/src/WpfMath/TexFontInfo.cs
+++ b/src/WpfMath/TexFontInfo.cs
@@ -1,4 +1,4 @@
-﻿using System;
+using System;
 using System.Collections.Generic;
 using System.Windows.Media;
 using WpfMath.Exceptions;
@@ -141,7 +141,7 @@ namespace WpfMath
 
         public double[] GetMetrics(char character)
         {
-            if (metrics.Length <= character)
+            if (metrics.Length <= character || metrics[character] == null)
             {
                 throw new TexCharacterMappingNotFoundException(
                     $"Cannot determine metrics for '{character}' character in font {FontId}");
--- a/src/WpfMath/TexFontInfo.cs
+++ b/src/WpfMath/TexFontInfo.cs
@@ -1,4 +1,4 @@
-using System;
+﻿using System;
 using System.Collections.Generic;
 using System.Windows.Media;
 using WpfMath.Exceptions;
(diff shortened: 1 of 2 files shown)


src/WpfMath/TexFontInfo.cs
```diff
@@ -1,4 +1,4 @@
-﻿using System;
+using System;
```
alexreg: What's going here? Seems like an identical line, hrmm.

sskodje: I have no idea..

alexreg:
Maybe just reset it then.

src/WpfMath/TexFontInfo.cs
sskodje: This is what git diff says.

@@ -1,4 +1,4 @@
-<EF><BB><BF>using System;
+using System;^M

I must admit not exactly sure what this means or how i fix it.

alexreg: Ah, it's a Windows-style line ending (`^M` is a carriage return). Maybe this will help: https://superuser.com/questions/194668/grep-to-find-files-that-contain-m-windows-carriage-return

sskodje: Yeah i figured it out, the file had been saved as UTF-8 instead of UTF-8-BOM. It should be fixed now.

ForNeVeR:
Thank you for the contribution.

I'm going to properly review that (and probably merge) tonight; ETA 12 hours.

As a side note: we do care about internationalization, it's just our font doesn't support most of the unicode characters. We want to eventually do something about it. For now, you could use `\text{æøå}` in your formula, it will fall back to Arial font (you may change it).

src/WpfMath/TexFontInfo.cs
ForNeVeR: Regarding BOM: we've [discussed these a while ago](https://github.com/ForNeVeR/wpf-math/pull/71#discussion_r105537740), but nobody was particularly interested, so I single-handedly decided to let the BOM stay as it is. Probably we'll need a small research regarding current behavior of compilers in case we want to remove it. Current "consensus" is that we use BOM (because it helps .NET compilers to work with Unicode in sources), and I'd prefer everyone (incl. myself) to not change that without further discussion.

ForNeVeR:
Again, sorry for the delay and thank you for the contribution! 👍

PR #103, opened 2018-01-21
Fixed a crash if inserting a whitespace after a _ or ^ symbol.

sskodje:
The ReadScript method is checking for position length before potentially incrementing it with SkipWhiteSpace, leading to index out of bounds exception if you insert a whitespace after a _ or ^. Moving SkipWhiteSpace before the length check seems to fix the issue.


alexreg:
Looks good to me. Thanks for the PR, and including a test! I'll let @ForNeVeR check over it quickly before merging though.

ForNeVeR:
Sorry for the delay and thanks for the contribution!

PR #113, opened 2018-02-03
Fix rendering for stuff like \text{æ,}

### LINKED ISSUE #104: Error when rendering `\text{æ,}`
Run the example program and enter the formula `\text{æ,}`.

It will for some reason fail with `NullReferenceException` in `DefaultTexFont.GetMetrix`.

ForNeVeR:
It turned out that the problem lies in the fact that WPF-Math tries to use `SymbolAtom`s in context of `\text` environment. `SymbolAtom` is tied to `TeXFont` and can't be used with `SystemFont`, so I decided to always use `CharAtom`s inside of `\text`.

- makes `CharAtom`s instead of `SymbolAtom`s inside `\text`
- marks `SystemFont` as not supporting the font metrics at the moment (so we won't try to query them in `RowAtom`)
- [x] **TODO**: raise an issue to support metrics (kerning and ligatures) in `SystemFont` → #114
- closes #104

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -5,4 +5,5 @@
 bin/
 obj/
 
+*.DotSettings
 *.user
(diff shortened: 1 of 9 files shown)


ForNeVeR:
Thanks for the review!

PR #116, opened 2018-02-23
Add StringSpan type to save the source information in Atoms and Boxes

yamamoto-reki:
Hello!

If there is no problem, please incorporate this code.
I have created basic code that conveys a range of source strings to Atoms and Boxes.
Currently only CharAtom and CharBox will work.
It is quiet to continue implementing other Atoms and Boxes.

Thanks.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -5,4 +5,5 @@
 bin/
 obj/
 
+*.DotSettings
 *.user
--- a/README.md
+++ b/README.md
@@ -31,7 +31,7 @@ The library was originally ported from the [JMathTex project][jmathtex], copyrig
 License Notes
 -------------
 
-The project code is licensed under the terms of [MIT license][license]. The original resources from [JMathTeX][jmathtex] (`DefaultTexFont.xml`, `GlueSettings.xml`, `PredefinedTexFormulas.xml`, `TexFormulaSettings.xml`, `TexSymbols.xml`) are taken from the [GPLv2-distributed][gpl] [JMathTeX][jmathtex], but JMathTeX authors have granted permission to redistribute these resourses under the MIT license.
+The project code is licensed under the terms of [MIT license][license]. The original resources from [JMathTeX][jmathtex] (`DefaultTexFont.xml`, `GlueSettings.xml`, `PredefinedTexFormulas.xml`, `TexFormulaSettings.xml`, `TexSymbols.xml`) are taken from the [GPLv2-distributed][gpl] [JMathTeX][jmathtex], but JMathTeX authors have granted permission to redistribute these resourses under the MIT license. See the [wiki][wiki-license-info] for details.
 
 The [fonts][] `cmex10.ttf`, `cmmi10.ttf`, `cmr10.ttf`, and `cmsy10.ttf` and `cmtt10.ttf` are under the [Knuth License][knuth-license].
 
@@ -48,6 +48,7 @@ The [fonts][] `cmex10.ttf`, `cmmi10.ttf`, `cmr10.ttf`, and `cmsy10.ttf` and `cmt
 [knuth-license]: http://ctan.org/license/knuth
 [launchpad]: https://launchpad.net/wpf-math
 [nuget]: https://www.nuget.org/packages/WpfMath/
+[wiki-license-info]: https://github.com/ForNeVeR/wpf-math/wiki/Additional-license-info
 
 [badge-appveyor]: https://ci.appveyor.com/api/projects/status/b26m3rpfcgb91gdg/branch/master?svg=true
 [badge-nuget]: https://img.shields.io/nuget/v/WpfMath.svg
--- a/README.md
+++ b/README.md
@@ -18,6 +18,11 @@ For a more detailed sample, check out the [example project][example]. It shows t
 
 ![Screenshot of example project](docs/example-screenshot.png)
 
+Documentation
+-------------
+
+- [How to prepare `DefaultTexFont.xml` from the font file][docs-prepare-font]
+
 History
 -------
 
@@ -30,6 +35,7 @@ The project code is licensed under the terms of [MIT license][license]. The orig
 
 The [fonts][] `cmex10.ttf`, `cmmi10.ttf`, `cmr10.ttf`, and `cmsy10.ttf` and `cmtt10.ttf` are under the [Knuth License][knuth-license].
 
+[docs-prepare-font]: docs/prepare-font.md
 [example]: WpfMath.Example/
 [fonts]: src/WpfMath/Fonts/
 [gpl]: docs/JMathTeX-license.txt
--- a/docs/prepare-font.md
+++ b/docs/prepare-font.md
@@ -1,0 +1,23 @@
+How to prepare XML from TTF
+===========================
+
+1. To gather font metrics:
+   ```console
+   $ ttf2tfm filename.ttf -v filename.vpl
+   $ tftopl filename.tfm > filename.tpl
+   ```
+   That will generate `filename.tfm` (binary) and `filename.vpl`, `filename.tpl`
+   (text).
+
+   `ttf2tfm` and `tftopl` are parts of the standard TeX distribution (at least
+   TeX Live).
+2. May be helpful:
+   ```console
+   $ ttx filename.ttf
+   ```
+
+   That will generate `filename.ttx` (XML).
+
+   [ttx][]
+
+[ttx]: https://github.com/fonttools/fonttools
--- a/docs/prepare-font.md
+++ b/docs/prepare-font.md
@@ -1,5 +1,17 @@
-How to prepare XML from TTF
-===========================
+How to prepare `DefaultTexFont.xml` from the font file
+======================================================
+
+This document describes the correspondence between the font files we use and
+data in `DefaultTexFont.xml`. The conrete approach (i.e. script files to
+regenerate XML data) hasn't been reproduced yet, although we were able to get
+nearly all the required information.
+
+Data in our XML files is often inaccurate and/or seemlingly generated from other
+TTF files than the files in the repository. Be prepared for that. When in doubt,
+refer to [the original TFM files][tfm] (they seem to correspond to our data
+better than the TFM files generated from our TTF files).
+
+There are two helpful toolsets to work with our TTF files:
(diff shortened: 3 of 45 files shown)


ForNeVeR:
Aha, I can see that the tests failed because they're trying to compare the `Source` field contents. I think we'll need to disable the comparer for that field; I'll take care of that.

yamamoto-reki:
I fixed it because StringSpan was not handed over to the end of the parser due to some mistakes.
Also, I modified the sample program to make it easier to understand new functions.
Try changing the selection range of the text box.

ForNeVeR:
I am sorry, some super important IRL stuff taking longer than I expected :(

I'll do my best to review the code in a couple of days.

yamamoto-reki:
When I actually implemented it, it became a bigger change than I expected. I'm sorry.
I am trying to fix unit tests. And I will create a new unit test.
Please wait for a few days.

ForNeVeR:
Don't worry, i'm still positive about these changes. I will help you to merge them properly (probably we'll split them into lesser PR's to ease review, I'll think about it). Hope we'll merge them!

For now, I'll add a `[WIP]` mark to the title of your PR. Please notify me when you're ready to merge that or in case you need help!

ForNeVeR:
_Also, I'm sorry about not commenting here earlier: I know I promised to review these in "a couple of days", but there're big (mostly positive) things happening in my life and I couldn't get to this PR. I'm not ignoring it and I will definitely get to is as soon as possible in my current situation._

ForNeVeR:
Alright, finally, I'm ready to do the review.

I can see that you're including stuff for formula editing here, too. I'd prefer to merge `StringSpan` separately. (I'd like to include formula editing stuff also, but later, when it'll be ready.)

I'll take a look and maybe create a new PR with only `StringSpan` parts, we'll see.

ForNeVeR:
### Summary of this PR:

- add `Atom.Source` and `Box.Source` properties, make sure they're filled correctly everywhere
- add `BoxWalker` and `FormulaUtils` classes to show the selected formula parts in the visual control
- add `Atom.Copy()` abstract method and implement it in all the derived classes. It seems like it's used [only in one place](https://github.com/yamamoto-reki/wpf-math/blob/e09551e5fe2fff0d5d387f9246e2dad0d6ca7d84/src/WpfMath/TexFormulaParser.cs#L485-L487)
  (we **have** to copy the Atom received from the `predefinedFormula` dictionary because it's a global dictionary and nobody should mutate these objects)

I think that we will eventually merge these changes. I can see technical reasons for the changes and overall I think that design is okay. Great work, @yamamoto-reki, thank you 👍 

Although I have some points I'd like to discuss with my co-maintainers. @alexreg, @gsomix, please help me to decide about them.

1. I think that `BoxWalker` and `FormulaUtils` are out of the scope of this PR. I'd like to review them separately later.
2. I don't like `Atom.Copy()` abstract method at all. It will be difficult to maintain when we add properties into our objects. I'd like to discuss the alternatives. Maybe turn `predefinedFormula` into a factory, like `Dictionary<string, Func<Atom>>`? What do you guys think?
3. I've changed my mind about the unit tests. I think we should add proper `Source` attributes to our test atoms and check them altogether with the other things. _(I can add them myself.)_
4. I don't think we want `Source` property on our `Box`es, because I'm not sure if we want to directly tie a visual formula tree to the textual source. Although I understand that the user may need that property. So, maybe we should add `SourceAtom` instead? And implement a `Source` as `StringSpan Source => SourceAtom?.Source`?
5. It looks like @yamamoto-reki is a fresh git user (and it's not a problem at all, any contributors are welcome! 🥇) and thus making many unnecessary merge commits. I'll manually clean them up before merging the code into our repository.
6. **An open question**: should we make a `Source` a mandatory property (e.g. require it in `Atom` constructor and check for `null`) or should we leave it optional (and thus lose any non-null guarantees)? I'm thinking about the cases when the user hasn't any textual sources, e.g. constructing a formula programmatically or reading an image (and not a text file).

ForNeVeR:
Regarding the `Atom.Source` property and whole `Atom.Copy` thing: my colleagues gave me an idea that we shouldn't extend our immutable `Atom` with the mutable properties, and should add a new class instead. Something like
```csharp
class AtomWithSource
{
    public Atom BaseAtom { get; set; }
    public StringSpan Source { get; set; }
}
```

alexreg:
> Regarding the Atom.Source property and whole Atom.Copy thing: my colleagues gave me an idea that we shouldn't extend our immutable Atom with the mutable properties, and should add a new class instead. Something like
> 
> ```csharp
> class AtomWithSource
> {
>     public Atom BaseAtom { get; set; }
>     public StringSpan Source { get; set; }
> }
> ```

I tend to agree with this. Nice separation of concerns.

alexreg:
> * I think that BoxWalker and FormulaUtils are out of the scope of this PR. I'd like to review them separately later.

Totally up to you. I don't mind either way, but it sounds fair.

> I don't like Atom.Copy() abstract method at all. It will be difficult to maintain when we add properties into our objects. I'd like to discuss the alternatives. Maybe turn predefinedFormula into a factory, like Dictionary<string, Func<Atom>>? What do you guys think?

Yes, it seems like a lot of work to maintain. If we were going to go this way, it would make more sense to make all the atom types value types (i.e. structs), so we get (cheap) copying for free. I haven't looked closely at the cons of doing this, but it might be nice to go with the immutable approach moving forwards. If not (or in addition), a builder for predefined formulae makes sense.

I also noticed that for properties the `this.` prefix is not being used. This is good practice, and convention throughout the existing codebase. We need to normalise this.

> I've changed my mind about the unit tests. I think we should add proper Source attributes to our test atoms and check them altogether with the other things. (I can add them myself.)

Yep, let's add them in separate tests. They can come after though.

> I don't think we want Source property on our Boxes, because I'm not sure if we want to directly tie a visual formula tree to the textual source. Although I understand that the user may need that property. So, maybe we should add SourceAtom property instead? And implement a Source as StringSpan Source => SourceAtom?.Source?

Agreed. Definitely shouldn't be on `Box`es.

> It looks like @yamamoto-reki is a fresh git user (and it's not a problem at all, any contributors are welcome! 🥇) and thus making many unnecessary merge commits. I'll manually clean them up before merging the code into our repository.

Git rebase to the rescue!

> An open question: should we make a Source a mandatory property (e.g. require it in Atom constructor and check for null) or should we leave it optional (and thus lose any non-null guarantees)? I'm thinking about the cases when the user hasn't any textual sources, e.g. constructing a formula programmatically or reading an image (and not a text file).

I think my previous reply answers this. (So no, it shouldn't be mandatory. Everything will be in terms of the base `Atom` class. You can then attempt to case an `Atom` as an `AtomWithSource` to see if it has source information attached.)

yamamoto-reki:
Thank you for reviewing.
I'd like to fix the problem pointed out early and well.
I agree with separating Atom and StringSpan.
And I will also work on other issues.
Perhaps it would be better to branch again from master and reimplement it.

ForNeVeR:
@alexreg 

> I also noticed that for properties the `this.` prefix is not being used. This is good practice, and convention throughout the existing codebase. We need to normalise this.

I disagree, let's discuss that. I think that there's guideline to avoid `this.` in some of the popular C# coding guidelines I'm personally trying to follow when writing in C#:

- [corefx guideline](https://github.com/dotnet/corefx/blob/master/Documentation/coding-guidelines/coding-style.md#c-coding-style) (see item 4)
- ReSharper default coding style

I can see that `this.` is more or less consistently used in the existing code base, although I try to remove any unnecessary code from the methods I change. `this.` falls into that category.

After all, in properly written code `this.` should almost never be necessary: when you see `var x = <something>`, it's usually unambiguous:
- either `<something>` starts with `_`, and then it's a field: `var x = _field`
- or it starts with a capital, and then it's a property (or a static readonly field per ReSharper guideline)

The only place when `this.` is "absolutely necessary" is when you're calling an extension method on it (and this is uncommon).

alexreg:
> I disagree, let's discuss that. I think that there's guideline to avoid this. in some of the popular C# coding guidelines I'm personally trying to follow when writing in C#:

The official VS (Roslyn) guideline recommends it, last time I checked. In any case, it makes code a lot more readable, since you can determine instance properties just by looking at them. Huge win. You also don't have to get involved in any shit like underscore prefixing, which plagues Java.

ForNeVeR:
Java doesn't generally use underscore prefixing, and that's the reason they prefer `this.`. I think you've actually mistook C# and Java here: in C#, people are often "plague" their code with underscores, and in Java they never do that :)

ForNeVeR:
Also, the [Roslyn repo](https://github.com/dotnet/roslyn/blob/master/CONTRIBUTING.md#coding-style) now references the same [style document](https://github.com/dotnet/corefx/blob/master/Documentation/coding-guidelines/coding-style.md) I linked before (the same that recommends against `this.`).

alexreg:
> Java doesn't generally use underscore prefixing, and that's the reason they prefer this.. I think you've actually mistook C# and Java here: in C#, people are often "plague" their code with underscores, and in Java they never do that :)

I've seen quite the opposite, and am quite sure of it! I guess we've had different experiences. Anyway, I think I've offered good rationale here; either way though, we need to be consistent about usage.

alexreg:
> Also, the Roslyn repo now references the same style document I linked before (the same that recommends against this.).

You're confusing the guidelines for Roslyn itself, and Roslyn style analysers, which insist on using `this.` for all properties (and possibly more, I forget).

ForNeVeR:
Alright, I agree. I'll extract all the messages about code style to a separate issue later. We'll establish a code style document and add a linter or style analyzer to check it for us. `this.` will be mandatory.

Alexander, thanks for discussion!

alexreg:
@ForNeVeR Okay, thanks for hearing me out! Sounds like a good plan for the future. We'll leave it for a separate issue, so we don't pollute this PR discussion any more. :-)

ForNeVeR:
Comments about the coding style were moved to #120. I will not explicitly impose any additional code style requirements to this PR yet.

ForNeVeR:
All the relevant changes from this PR were merged. @yamamoto-reki thanks a lot for your hard work; your changes were very useful for our project! 👍

PR #122, opened 2018-03-25
Fix Shift of ScriptsAtom without any scripts

### LINKED ISSUE #119: Big operator rendering problem inside `\frac`
Hello,

First I would like to say WpfMath is a wonderful library. I had searched something like this for a long time.

I went into a small rendering bug when using this kind of formula: \frac{1}{\sum{a}}
The "a" is not at the right place but looks like a superscript. It happens only when the sum is under a fraction.

Best regards,
Pierre

ForNeVeR:
I've found that `ScriptsAtom` usually returns a shifted `CharAtom` (e.g. a `CharAtom` whose `Shift` is set), but in case when both `SubscriptAtom` and `SuperscriptAtom` are `null`, it returns essentially just a `new CharAtom` without the `Shift`. It causes bugs such as #119. I've decided to make `ScriptsAtom` always return the same kind of `CharAtoms` (i.e. the shifted ones).

Closes #119.

### DIFF
--- a/src/WpfMath.Tests/BoxTests.fs
+++ b/src/WpfMath.Tests/BoxTests.fs
@@ -32,3 +32,15 @@ let ``Box for \text{æ,} should be created successfully``() =
     let atom = parse @"\text{æ,}"
     let box = atom.CreateBox(environment)
     Assert.NotNull(box)
+
+[<Fact>]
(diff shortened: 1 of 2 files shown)


ForNeVeR:
Thanks for review!

PR #123, opened 2018-04-01
SourceSpan support

### LINKED ISSUE #115: Give a range of source string to Box
Hello!

Can you give range information of source string to Box?
Because it is very convenient when making formula editor.

Thanks.

ForNeVeR:
Closes #115.

This is a rebased version of PR #116. I will be fixing the tests here and polishing the changes.

### TODO

- [x] Rebase all the remaining changes
- [x] Fix the issues mentioned in https://github.com/ForNeVeR/wpf-math/pull/116#issuecomment-373978841 and further discussion:
  - [x] Remove `BoxWalker` and `FormulaUtils`
  - [x] Drop `Atom.Copy` in favor of `AtomWithSource`
  - [x] ~~Add `Box.SourceAtom`, drop `Box.Source`~~ _I decided to not do that_
- [x] Change `StringSpan` into `SourceSpan`
- [x] Add tests for `SourceSpan`s
- [x] Fix the differences between #116 and this PR (see my comment below)


ForNeVeR:
Well, I've started removing `Atom.Copy` in favor of `AtomWithSource` and I have a feeling that something's wrong. Look: we have all these references all over the place (e.g. `BigOperatorAtom`) referencing its' `BaseAtom` and it looks like we should give this property a signature `AtomWithSource BaseAtom { get; set; }`

Our code will become a mess with slices of `Atoms` and `AtomWithSource`s, I don't like that.

Looks like we should return back to the design board. Probably we will need that `Atom.Copy` after all? Maybe other guys' suggestion about rewriting everything in F# wasn't so radical as I thought?

Because, well, current problem is that we cannot get cheap class copying in C#, so we have to implement this `Copy` method in every `Atom` descendant, and I extremely dislike that. What could we do? Make everything `struct`? Will it work? Any other suggestions?

alexreg:
@ForNeVeR As long as there are no many-to-one or one-to-many references (from memory I don't think so), then converting everything to value types (i.e. structs) would seem to make sense to me. You get efficient copying for free then.

ForNeVeR:
Yeah, working on it in a separate branch. We'll see what else we can do. I have some other ideas.

ForNeVeR:
>   - [ ] Add `Box.SourceAtom`, drop `Box.Source`

Now I'm not sure about that. Currently, all of our `Atom`s are internal classes, so we cannot publish them through a public member. And I'd keep it as is for the moment, because the `Atom` API will likely be unstable in the future.

We could mark only the root of the `Atom` type hierarchy as `public`, but I can't see a point for that: the only usable public members would be `TexAtomType` and `SourceSpan`. Basic `Atom` is useless if the user cannot see its' structure.

Hereby I revoke my requirement about adding `Box.SourceAtom`. For now, let's leave a `Box.Source` member intact.

ForNeVeR:
Also, we have some bugs here. @yamamoto-reki made the selected blocks to highlight in the example project. E.g.

image
(the highlighting works well and it's accurate enough)

In my PR, it doesn't highlight some blocks and sometimes behaves weird:
image

Before merging, we'll need to fix that weirdness and make sure everything works well.

src/WpfMath/Controls/FormulaControl.xaml.cs
```diff
@@ -75,6 +94,18 @@ public ControlTemplate ErrorTemplate
             "ErrorTemplate", typeof(ControlTemplate), typeof(FormulaControl),
             new PropertyMetadata(new ControlTemplate()));
 
+        public static readonly DependencyProperty SelectionStartProperty = DependencyProperty.Register(
+            "SelectionStart", typeof(int), typeof(FormulaControl),
```
ForNeVeR: Don't forget to use `nameof` here!

src/WpfMath/TexFormula.cs
```diff
@@ -49,18 +49,23 @@ public TexRenderer GetRenderer(TexStyle style, double scale, string systemTextFo
             return new TexRenderer(CreateBox(environment), scale);
         }
 
-        public void Add(TexFormula formula)
+        public void Add(TexFormula formula, SourceSpan source = null)
```
ForNeVeR: The binary compatibility has been broken here, unfortunately (the source compatibility was preserved). We could fix that. Should we?

src/WpfMath/TexPredefinedFormulaParser.cs
```diff
@@ -1,341 +1,339 @@
-using System;
```
ForNeVeR: Previously this file had `CRLF` and now it has `LF` (it was replaced because I renamed it to match the class name). I think that's alright.

Please check the online diff with `?w=1` appended to the page URL to see the real changes (it will ignore the whitespace diff): https://github.com/ForNeVeR/wpf-math/pull/123/files?w=1#diff-e81f0c498a5c8d83d49d999c0375f718

src/WpfMath/TexFormula.cs
ForNeVeR: Don't think so. Hopefully nobody uses that API. And we're still in API-breaking period of our semver (`0.x` versions).

ForNeVeR:
Alright, it's ready for review, guys.

@yamamoto-reki could you also please take a look?

My next step will be to review (and probably merge) the remaining changes (`BoxWalker`, `FormulaUtils`) from your PR #116.

src/WpfMath/TexFormulaHelper.cs
```diff
@@ -1,34 +1,30 @@
...
-        public TexFormula Formula
-        {
-            get;
-            private set;
-        }
+        public TexFormula Formula { get; }
 
+        // TODO[F]: Review the cases where the formula gets constructed from this.Formula.RootAtom wrapped in something
+        // (e.g. SetFixedTypes): in these cases, it looks like we should clean the SourceSpan inside of a formula and
+        // set it for the new root atom only?
```
ForNeVeR: Please file an issue about that!

ForNeVeR:
Thanks for the review!

PR #130, opened 2018-04-18
Summation two cyrillic symbols (#129)

TheOneAmir:
The problem is occurring because even though the Cyrillic symbols are inside "\Text{}", they are considered as SystemFont instead of DefaultTexFont by RowAtom.cs. As a result, the function to GetFontMetrics() inside SystemFont.cs is never called:
Function not called: 
<code>
        private TexFontMetrics GetFontMetrics(char c, Typeface typeface)
        {
            var formattedText = new FormattedText(c.ToString(),
                CultureInfo.CurrentUICulture,
                FlowDirection.LeftToRight,
                typeface,
                1.0,
                Brushes.Black);
            return new TexFontMetrics(formattedText.Width, formattedText.Height, 0.0, formattedText.Width, 1.0);
        }
</code>

By adding the following check inside RowAtom.cs, we check if the Element.TextStyle is actually "text" so we actually use SystemFont.cs for rendering the character. 

References issue: https://github.com/ForNeVeR/wpf-math/issues/129

### DIFF
--- a/src/WpfMath/RowAtom.cs
+++ b/src/WpfMath/RowAtom.cs
@@ -1,4 +1,4 @@
-﻿using System.Collections;
+using System.Collections;
 using System.Collections.Generic;
 using System.Collections.ObjectModel;
 using System.Linq;
@@ -114,7 +114,8 @@ namespace WpfMath
 
                 // Check if atom is part of ligature or should be kerned.
                 var kern = 0d;
-                if (hasNextAtom && curAtom.GetRightType() == TexAtomType.Ordinary && curAtom.IsCharSymbol)
+                if ((hasNextAtom && curAtom.GetRightType() == TexAtomType.Ordinary && curAtom.IsCharSymbol) &&
+                    !(this.Elements[i].GetType() == typeof(CharAtom) && ((CharAtom)this.Elements[i]).TextStyle == "text"))
                 {
                     if (nextAtom is CharSymbol cs && ligatureKernChangeSet[(int)nextAtom.GetLeftType()])
                     {


ForNeVeR:
Thanks for your contribution!

This solution fixes the issue, but I don't like the resulting code. I'll push my solution to your branch in about ≈24h, and we'll review it then. (Maybe my solution is more complex than yours, we'll discuss with other maintainers and decide what to apply.)

TheOneAmir:
Sure, sounds good to me. To be honest, I wasn't entirely sure of where the best place for the solution would be. It is possible to extend DefaultTexFont to handle the problem there as well but this didn't "feel" right. I look forward to seeing your solution!

ForNeVeR:
Alright, it's ready for review.

My main idea was to add `CharSymbol.IsSupportedByFont` API and use it in `RowAtom`. Although, I was forced to further pollute the `ITeXFont` with all these `bool assert = true` flags (they tell the `ITeXFont` whether it should throw exceptions or just ignore the errors and return `null`.

Other idea was to catch `TexCharacterMappingNotFoundException` etc. in my new `Supports<…>` methods, but I dislike it even more.

Dear reviewers, please take a look at the [last commit](https://github.com/ForNeVeR/wpf-math/pull/130/commits/f015a4ed05bc2d3d6dd1ecbbf7baafc28f388214): it pretty much describes differences between the initial idea of @TheOneAmir and my further so-called "improvements". Despite their ugliness, I still believe that approach is better than the initial one, because it is

1) offers more semantic by explicitly calling the stuff `IsSupportedByFont` instead of relying on some obscure flags such as `!(this.Elements[i].GetType() == typeof(CharAtom) && ((CharAtom)this.Elements[i]).TextStyle == "text")`
2) it seems to work better in some of the (imagined) corner cases when the `!(this.Elements[i].GetType() == typeof(CharAtom) && ((CharAtom)this.Elements[i]).TextStyle == "text")` is `true` but the font still supports the character and so we want to render it properly.

@TheOneAmir, please review that, too, if you're interested.

ForNeVeR:
We decided to take another look at the code. Current state is unsatisfactory.

TheOneAmir:
Right, so we're not going to use either solution? I can take a look and try to find another "cleaner" one.

ForNeVeR:
@TheOneAmir you're right. It's better to wait a day or two before I'll finalize my ideas in the code. I think it'll be much cleaner.

My current plan is:

1. Add some tests for `DefaultTexFont.GetChar` overloads and fix the errors there (yep, I've already find a couple of bugs).
2. Add [CSharpFunctionalExtensions](https://github.com/vkhorikov/CSharpFunctionalExtensions) package to our project (seems to be compatible with .NET 4.0, wonderful).
3. Change the current `GetChar` overloads to return `Result.Ok` / `Result.Error`, drop the `bool assert` flag.
4. Make the callers to analyze the `Result` returned and either rethrow the exceptions or just do their thing. Drop my new `IsSupported` methods, we don't need them.

PR #134, opened 2018-04-29
Add BoxWalker and FormulaUtils

ForNeVeR:
This is the second part of @yamamoto-reki's work on the project (extracted from #116).

It introduces two public classes, `FormulaUtils` and `BoxWalker`, I guess, to simplify the rendering code and to provide these tools to a library user.

Closes #116.


ForNeVeR:
You know, honestly I don't think we need these changes. It looks like the current code in `FormulaControl` does its' job well, and I see no immediate reason to merge these changes right now.

Other opinions, guys?

ForNeVeR:
Alright, for now I decide to not merge that. @yamamoto-reki if you still need these changes, feel free to notify me. We will definitely find a solution!

PR #135, opened 2018-05-21
Move atoms and boxes to namespaces

### LINKED ISSUE #124: Move all the Atoms and Boxes to the corresponding namespaces
Currently, we have all our stuff in one namespace and one directory. It interferes a bit with global refactorings: non-important stuff is constantly on my way when I'm looking for particular `Atom` or `Box`.

I suggest we move all the `Atom` inheritors to `WpfMath.Atoms` namespace and all `Box` inheritors to `WpfMath.Boxes` namespace.

**Please, do it only after #116 and #123 get sorted up!**

ForNeVeR:
Closes #124.

We have a small breaking change: public class `Box` has been moved to a new namespace. I think it's okay (because we're still in a `0.x` semver sector where we're supposed to break things, although I'd like to release 1.0 soon — and stop breaking things then).

Also, I've accidentally replaced Windows line endings (`\r\n`) in some of the files with cross-platform line-endings (`\n`). As a repository maintainer, I think that's a good change.

My refactoring tools also accidentally added a couple of `this` references according to our coding style — I suppose it's also a good thing.

My points aren't final, so, if anyone wants to discuss the changes (and wants me to put `Box` back to avoid API breaking or to put `\r\n` back) — let's discuss!


ForNeVeR:
Note for reviewers: when reviewing line ending conversion, please use the new GitHub feature to disable whitespace diffs. Here's how it looks on my screen:

image

alexreg:
I’m happy with these conventions. 👍🏻

ForNeVeR:
Thank you for the review, guys!

PR #137, opened 2018-05-26
Use .NET Core SDK

### LINKED ISSUE #136: Migrate to .NET Core SDK
I'd like WPF-Math to be migrated to .NET Core SDK (the so-called "new SDK", the one released with .NET Core and controlled mainly via the console `dotnet` command).

Pros:
- we'll be ready for cross-platform builds
- ⚠ our `csproj` files will be much more concise
- ⚠ no more `packages.config` (`<PackageReference>` is now supported by SDK)
- better modern tooling support (e.g. `dotnet sln` console command etc.)
- standard project build command (`dotnet build` instead of cryptic msbuild+nuget combinations)
- ⚠ probably more straightforward `dotnet pack` experience
- ⚠ still supports multiple frameworks in various combinations (e.g. `netstandard+net40` for portable libraries)

Cons:
- WPF is not a first-class citizen in a modern SDK world
- user experience of working with the project may be degraded

Degraded user experience is not acceptable for us, so we'll need to check if everything works correctly after migration (focus on scenarios such as adding a new user control into example project: they are more likely to cause issues). If the user experience will still be negative, we won't migrate.

I've found the following documentation about migrating WPF projects to new SDK:
- https://stackoverflow.com/a/44539088 (some boilerplate in the `csproj` file, but should work okay; need to check with Visual Studio)
- https://github.com/SunburstApps/MSBuildSdks/blob/master/src/Sunburst.NET.Sdk.WPF/README.md (`csproj` will be even more clean, but still need to check if VS tooling won't fail us)

ForNeVeR:
Closes #136.

TODO:
- [x] test user experience in VS
- [x] update packaging instructions
- [x] fix `appveyor.yml` (get rid of it completely? it seems we're satisfied by the defaults for now)

Fun fact: I've found that there was a `TexExpressionVisitor` class that wasn't included into project (the new SDK helped me to find that). I've deleted it because it obviously has no purpose.


ForNeVeR:
Please ignore the commits that reference this PR, #137 in the commit log: these were pushed accidentally and already cleaned up by my following force-push. I couldn't remove them from this GitHub page though. They are absent from my PR branch.

ForNeVeR:
Okay, ready for review!

ForNeVeR:
Hereby I approve these changes. Please report any issues you have with the new project system, I'll keep monitoring them.

PR #138, opened 2018-06-29
Client defined color parsing

B3zaleel:
Added support for parsing colors defined by the client using the old "color"(changed to "fgcolor" and "bgcolor") and colorbox commands.

PR #142, opened 2018-07-16
Color support

### LINKED ISSUE #165: Proposal for color format
While working on #142, I've checked internet for documentation about how other LaTeX-related tools deal with colors. I have not found any evidence that LaTeX supports direct color definition (e.g. `\color{rgb}{1.0,1.5,1.0}{mytext}`) for either `\color` or `\colorbox`. Although, I've found a [useful piece of documentation](https://en.wikibooks.org/wiki/LaTeX/Colors#Color_Models) on Wikibooks.

I don't want to add ability to **define** new colors in WPF-Math LaTeX documents, so I don't want to add  commands such as `\definecolor`. I believe that the markup should stay static at the moment and should not have any influence on itself (i.e. no macro or color or whatever definitions in markup).

So I suggest that we document and add a modified syntax for `\color` and `\colorbox` commands by adding an optional `mode` parameter. Full syntax will be:

```
\command [mode] {color} {text}
```

Where:
- `command` is either `color` or `colorbox`
- `mode` is one of `gray`, `rgb`, `RGB`, `HTML`, `cmyk` (for now let's stop on that, but the code should be robust enough to add more modes: xcolor manual includes some additional modes)
- `color` is either a predefined color name (as implemented already) if the `mode` argument wasn't provided or a proper color definition according to the wikibooks documentation
- additionally to the standard `color` definitions, we will support a transparency parameter that will be **optional** and should be added in a convenient manner to every mode we support

_The motivation for adding transparency is clear: while LaTeX was designed to provide mostly a printed documents, we know that we're working with display graphics in WPF-Math, and transparency could be useful and shouldn't be too hard to implement, so I think we should add it._

Here're some syntax samples:

```
\colorbox{red}{text} // already implemented, "red" is a predefined color name
\colorbox{red,0.5}{text} // predefined color with transparency
\colorbox[gray]{0.5}{text}
\colorbox[gray]{0.5, 0.5}{text} // added transparency; also note that embedded spaces should be supported
\colorbox[rgb]{0,0,0}{text}
\colorbox[rgb]{0,0,0, 0.5}{text}
\colorbox[RGB]{255,255,255}{text}
\colorbox[RGB]{255,255,255,0}{text}
\colorbox[HTML]{FFFFFF}{text}
\colorbox[HTML]{ffffff00}{text}
\colorbox[cmyk]{0,0,0,1}{text}
\colorbox[cmyk]{0,0,0,1,0.5}{text}
```

ForNeVeR:
Changes extracted from #138 by @B3zaleel.

Closes #165.

**TODO**:
- [x] leave only the original commands (`\color` and `\colorbox`)
- [x] fix the failing unit tests
- [x] think about the accepted color formats
  - [x] check if any existing TeX distributions have something like that
- [x] write the documentation for the new color input
- [x] add tests for the new input
- [x] fix the implementation according to the tests
- [x] make sure it's robust enough so the new color parsers could be added easily
- [x] review all the comments below
- [x] add tests for double/integer parse errors in the color parsers (all should throw `TexParseException`s)
- [x] add tests for negative / too big values for all the numbered models (should also throw `TexParseException`s)
- [x] make sure the corresponding infrastructure and helper classes are public
- [x] review all the code one final time


src/WpfMath/UserDefinedColorParser.cs
```diff
@@ -0,0 +1,264 @@
...
+                        resultColor.A = 255;
+                        resultColor.R = channelbytes[0];
+                        resultColor.G = channelbytes[1];
+                        resultColor.B = channelbytes[2];
+                        break;
+                    }
+                case ColorStringTypes.None:
+                    throw new TexParseException($"The text: {input} could not be parsed.");
+                default:
+                    break;
```
ForNeVeR: I would `throw` in a `default` case, because it's a situation that should never happen.

src/WpfMath/UserDefinedColorParser.cs
```diff
@@ -0,0 +1,264 @@
...
+        #region Color Parsing Helpers
+        /// <summary>
+        /// Specifies the type of <see cref="Color"/> the string might represent.
+        /// </summary>
+        private enum ColorStringTypes
+        {
+            /// <summary>
+            /// Type is a <see cref="string"/> consisting of byte values of the color.
+            /// </summary>
+            Byte_longString,
```
ForNeVeR: Could you please fix the naming here? Underscore in an enum item name seems odd. Please make that `ByteLongString` etc.

Also I recommend to rename these to something like `CommaSeparatedDecimal`, `PrefixedHex` etc.

src/WpfMath/UserDefinedColorParser.cs
```diff
@@ -0,0 +1,264 @@
...
+        }
+
+        /// <summary>
+        /// Gets the type of <see cref="ColorStringTypes"/> from the <paramref name="input"/>.
+        /// </summary>
+        /// <param name="input"></param>
+        /// <returns></returns>
+        private static ColorStringTypes GetColorString(string input)
+        {
+            if (IsByteHexTrain(input,6)==true)
```
ForNeVeR: You don't need `== true` in this method.

ForNeVeR: These numbers, 6, 8, 3 and 4, seems a little bit magical here. Could we have them in named constants?

src/WpfMath/UserDefinedColorParser.cs
```diff
@@ -0,0 +1,264 @@
...
+        /// 23,78,56=>true
+        /// 789,fp=>false
+        /// </example>
+        /// <param name="input"></param>
+        /// <param name="num"></param>
+        /// <returns></returns>
+        private static bool IsByteTrain(string input, int num)
+        {
+            bool StrCheck = false;
+            string[] arrStr = input.Trim().Split(',');
```
ForNeVeR: No need to `Trim` here: the input is already trimmed.

src/WpfMath/UserDefinedColorParser.cs
```diff
@@ -0,0 +1,264 @@
...
+        /// Returns a value that tells if the <paramref name="input"/> contains byte values, separated by a ",".
+        /// </summary>
+        /// <example>
+        /// 23,78,56=>true
+        /// 789,fp=>false
+        /// </example>
+        /// <param name="input"></param>
+        /// <param name="num"></param>
+        /// <returns></returns>
+        private static bool IsByteTrain(string input, int num)
```
ForNeVeR: 1. This check is a little bit faulty. It seems like it will succeed e.g. for string `"0,0,quack,0,0"` for `num = 4`. Please fix that!
2. I think that this whole method could be replaced with a concise one-liner. Will you try to do it youself or do you need my help with that?

src/WpfMath/UserDefinedColorParser.cs
```diff
@@ -0,0 +1,264 @@
...
+        {
+            bool StrCheck = false;
+            string[] arrStr = input.Trim().Split(',');
+            if (num == arrStr.Length)
+            {
+                int i = 0;
+                foreach (var item in arrStr)
+                {
+                    if (Byte.TryParse(item, out byte result) == true)
```
ForNeVeR: Please don't forget for `InvariantCulture` here!

src/WpfMath/UserDefinedColorParser.cs
```diff
@@ -0,0 +1,264 @@
...
+            /// <summary>
+            /// Type is a <see cref="string"/> consisting of byte values of the color.
+            /// </summary>
+            Byte_longString,
+            /// <summary>
+            /// Type is a <see cref="string"/> consisting of byte values of the color.<para/>
+            /// Alpha channel is strictly = 255.
+            /// </summary>
+            Byte_shortString,
+            //FloatString,----->Will include this later.
```
ForNeVeR: Please add a `TODO:` prefix for stuff like that. I usually file new issues about these TODOs when merging a PR.

src/WpfMath/UserDefinedColorParser.cs
```diff
@@ -0,0 +1,264 @@
...
+        /// Returns a value that tells if the <paramref name="input"/> contains byte hex values,
+        /// </summary>
+        /// <param name="input"></param>
+        /// <param name="num"></param>
+        /// <returns></returns>
+        /// <remarks><paramref name="input"/> should be left in the raw state(e.g.;#56e245, not 56e245).</remarks>
+        private static bool IsByteHexTrain(string input, int num)
+        {
+            bool StrCheck = false;
+            string subStr = input.Substring(1);
```
ForNeVeR: Empty string will fail here!

src/WpfMath/UserDefinedColorParser.cs
```diff
@@ -0,0 +1,264 @@
...
+            if (num == subStr.Length)
+            {
+                int c = 0;
+                for (int i = 1; i < num; i+=2)
+                {
+                    string item = subStr[i - 1].ToString() + subStr[1].ToString();
```
ForNeVeR: `subStr.Substring(i - 1, 2)` should do the trick better.

src/WpfMath/UserDefinedColorParser.cs
```diff
@@ -0,0 +1,264 @@
...
+                    if (Byte.TryParse(item, System.Globalization.NumberStyles.AllowHexSpecifier, CultureInfo.InvariantCulture, out byte result) == true)
+                    {
+                        c+=2;
```
ForNeVeR: It seems to me like this approach will consider cases like `33XX3333` as valid. Please fix it.

src/WpfMath/UserDefinedColorParser.cs
```diff
@@ -0,0 +1,264 @@
...
+            return StrCheck;
+        }
+
+        /// <summary>
+        /// Gets the <see cref="byte"/> values from the <paramref name="input"/> if the number of items are ==<paramref name="num"/>.
+        /// </summary>
+        /// <param name="input"></param>
+        /// <param name="num"></param>
+        /// <returns></returns>
+        private static byte[] ByteStringValues(string input,int num)
```
ForNeVeR: I don't like the separation of `IsByteTrain` and `ByteStringValues`, the same goes for ther hex counterparts. Could we make two functions here: one for "bytes" and one for "hex"? Let it do validation and extraction in one pass.

src/WpfMath/TexFormulaParser.cs
```diff
@@ -14,7 +14,7 @@ namespace WpfMath
     // TODO: Use TextReader for lexing.
     public class TexFormulaParser
     {
-        // Special characters for parsing
+        #region Special characters for parsing
```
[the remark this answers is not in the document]
B3zaleel: What about the "Summary" for classes and constructors rather than comments?

src/WpfMath/UserDefinedColorParser.cs
B3zaleel: I tried #33XX3333  but it doesn't work

src/WpfMath/TexFormulaParser.cs
```diff
@@ -393,20 +395,58 @@ private TexFormula ReadScript(TexFormula formula, SourceSpan value, ref int posi
...
-
-                case "color":
+                    }
+                case "fgcolor":
                     {
+                        //Command to change the foreground color 
                         var colorName = ReadGroup(formula, value, ref position, leftGroupChar, rightGroupChar);
-                        var remainingString = value.Segment(position);
+
+                        string remainingString = value.Substring(position);
```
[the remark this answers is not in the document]
B3zaleel: I haven't been able to work with the SourceSpan and Result structure because I'm stilk using VS 15.6.7 and they through a lot of errors when I try to build it.

src/WpfMath/UserDefinedColorParser.cs
B3zaleel: I've done that in the new branch

src/WpfMath/TexFormulaParser.cs
```diff
@@ -80,7 +80,8 @@ private static void Initialize()
                 "left",
                 "right",
                 "sqrt",
-                "color",
+                "fgcolor",
```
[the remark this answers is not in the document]
Happypig375: @ForNeVeR [`\color` and `\colorbox` are standard LaTeX](https://www.sharelatex.com/learn/Using_colours_in_LaTeX) while `\bgcolor` and `\fgcolor`'s origins are unknown. I suggest thinking twice if you are going to deprecate them.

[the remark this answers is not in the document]
ForNeVeR: Alright, @B3zaleel, I'd like to discuss the naming with you. I don't think we should replace widely known TeX commands with the new ones.

src/WpfMath/TexFormulaParser.cs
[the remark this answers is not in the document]
ForNeVeR: Yeah, `<summary>` would be okay; though you can't put the summary for the code block. It's a class- or method-level thing.

src/WpfMath/TexFormulaParser.cs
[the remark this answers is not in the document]
ForNeVeR: Well that's a pity. I'm not sure how to proceed. In this project (and generally in open-source) I tend to touch all the shiny new things. because it's interesting and educational experience for all the members involved. "Having fun" is the main driving force for my open-source contributions (while "creating good software packages and providing them to the users" is also a goal, it's not the driving force). Thus I don't think that falling back to ol' good C# 7.0 is an option for us.

Although I fully understand that there're users whose work environment enforces some strict policies and that updating the Visual Studio isn't an option in such environments.

I'm not sure how to proceed. Could you maybe set up your other machine with a more recent version of Visual Studio Community edition (it is enough for developing open source software)? Or even try to set it up using the Visual Studio Build Tools package and VSCode? _(Not sure if it includes the WPF build tools though, likely it isn't.)_

src/WpfMath/UserDefinedColorParser.cs
ForNeVeR: That branch is named `Feature-Hex-codes`, right?

src/WpfMath/TexFormulaParser.cs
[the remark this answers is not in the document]
ForNeVeR: Also you should look at the build server checks at the bottom of that page: e.g. [here](https://ci.appveyor.com/project/ForNeVeR/wpf-math/build/1.0.453) it complains about SourceSpan.

src/WpfMath/UserDefinedColorParser.cs
B3zaleel: Yes

src/WpfMath/TexFormulaParser.cs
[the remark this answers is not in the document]
B3zaleel: What if they could coexist, since they would do the same thing?

src/WpfMath/TexFormulaParser.cs
```diff
@@ -393,20 +395,58 @@ private TexFormula ReadScript(TexFormula formula, SourceSpan value, ref int posi
...
+                            return new StyledAtom(remaining.RootAtom, null, new SolidColorBrush(color));
+                        }
+                        else
+                        {
+                            try
+                            {
+                                Color color1 = UserDefinedColorParser.ParseUserColor(colorName);
+                                return new StyledAtom(remaining.RootAtom, null, new SolidColorBrush(color1));
+                            }
+                            catch
```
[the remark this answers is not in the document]
B3zaleel: I may not be able to change the code now(it would take a lot of time), since my internet service is poor and Visual Studio is massive. I would really appreciate it if you could make the needed changes.

src/WpfMath/UserDefinedColorParser.cs
B3zaleel: The catch block handles that

src/WpfMath/TexFormulaParser.cs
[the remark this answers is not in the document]
ForNeVeR: Honestly I can't se  reason for us to add new commands except from widely-recognized ones. Could you just add your new behavior to the old commands?

src/WpfMath/TexFormulaParser.cs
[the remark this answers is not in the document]
ForNeVeR: Not a problem, I could fix that.

src/WpfMath/UserDefinedColorParser.cs
ForNeVeR: Well, "catch-all" clause could catch chat. But it's a widely recognized antipattern, we definitely don't want to catch everything in our code!

src/WpfMath/TexFormulaParser.cs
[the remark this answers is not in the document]
B3zaleel: Thank you very much

src/WpfMath/TexFormulaParser.cs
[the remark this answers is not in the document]
B3zaleel: That would be fine

src/WpfMath/TexFormulaParser.cs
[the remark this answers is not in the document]
ForNeVeR: Okay, I've fixed the compilation issue.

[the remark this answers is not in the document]
B3zaleel: Thank you

src/WpfMath/UserDefinedColorParser.cs
ForNeVeR: Okay, you're right. I haven't noticed the check `if (num == subStr.Length)`.

ForNeVeR:
Alright, I'm going to finalize this task based on @B3zaleel's contributions. Stay tuned.

ForNeVeR:
Thanks, everyone! Now this is (finally!) ready for merge.

Issue #128, opened 2018-04-15
\colorbox renders root index invisible


ForNeVeR:
This is okay: 
![image](https://user-images.githubusercontent.com/92793/38775559-bfe070ec-40af-11e8-8481-bca852b0d565.png)

But this is not (`\sqrt{2}` doesn't show `2` anymore):
![image](https://user-images.githubusercontent.com/92793/38775565-d8056dda-40af-11e8-97a2-5c789b63d646.png)

The same problem could be encountered with the new formula box selection mode (see #123): e.g. the root index `\sqrt[4]` isn't visible when the formula part is selected:
![image](https://user-images.githubusercontent.com/92793/38775575-1b7dcc92-40b0-11e8-917b-c80d6387a973.png)

B3zaleel:
That happened because the radical was visually in front of the root, so i had to reorder the arrangement in the *Radical Atom*. [Here it is](https://github.com/B3zaleel/wpf-math/blob/master/src/WpfMath/Atoms/Radical.cs)

ForNeVeR:
@B3zaleel I'm sorry but your contributions are getting increasingly harder to follow. I would appreciate very much if you stop pushing everything to your `master` branch. Please 😭😭😭

B3zaleel:
Ok, sorry about that

B3zaleel:
I just fixed an arrangement issue with the *Radical*, should I create a new branch for that? Even though it's  like 2 SLOCs?

ForNeVeR:
Yes, please. Create a branch from the current `master` from this repository and send a separate PR.

We haven't merged anything from your repository yet for this exact reason: you have tons and tons of changes there, and nobody had the time to review all of that.

On the other hand, small accurate PR could be reviewed and merged almost instantly.

Thanks!

ForNeVeR:
Alright, the issue here is that the root index gets drawn first, and then it gets overdrawn by the colored background of the whole radical sign and everything.

We should either rearrange the items in the radical atom's box, or change the background rendering strategy (that it only set the background once per atom, and don't transitively touch all the atoms inside of a child one).

ForNeVeR:
Currently, the best idea seems to be to create a new container `LayeredBox` that will place one box directly on top of another. This should help us to remove the negative-sized-atom hackery currently placed in the `Radical` atom, and also will allow to set the proper rendering order for the subboxes while keeping the background rendering code.

ForNeVeR:
[...]

PR #146, opened 2018-08-15
#128 Root index visibility

B3zaleel:
Added some changes to fix the root index visibility.


B3zaleel:
Now it is much stable than before, please try it out.

ForNeVeR:
And why have you closed this? 😢

ForNeVeR:
This PR makes the radicals look worse (left: old, right: new):
image

PR #147, opened 2018-08-15
Added a file for logging errors

B3zaleel:


ForNeVeR:
I don't understand what's the use of this class. It is `public`, so it's probably indended for the library users.

I can't see how we could help our users by providing yet another (probably imperfect!) logger implementation.

@B3zaleel could you please explain?

alexreg:
I agree with @ForNeVeR on this. Logging isn't really something that should even be part of a library like this, in my view, let alone our own implementation.

Happypig375:
And it is buggy. Why is the `LoggedErrors` field `static`?

ForNeVeR:
Closed with #148.

PR #148, opened 2018-08-15
Implemented the error logger

B3zaleel:


ForNeVeR:
Alright, now I can see the reason for #147 (sorry, I wasn't looking at #148 when checking #147).

But I can't see the reason of this PR. What's the purpose?

The usual workflow is: start WPF-Math in a debug mode (likely in IDE), catch any errors with your debugger and examine them right at the start.

Why log the errors separately? Are we sure we need this?

/cc @gsomix

ForNeVeR:
Well, no, we're not going to expose that logger to our end users.

In case a user need to examine errors from our Formula control, they always could check the corresponding `Errors` property and redirect these errors to their logging system of choice.

For now, I'm going to close this PR.

Although, feel free to argue, I'm always open for discussion. Maybe you think we need this for our internal usage (e.g. move that into `WpfMath.Example`)?

alexreg:
If errors are encountered while parsing a TeX formula or whatnot, they should be reported via a data structure or `Exception`, not an error logger.

PR #150, opened 2018-08-20
#60 simplified command arguments

### LINKED ISSUE #60: Command arguments passed without braces
Currently `\frac22` fails with an error "missing {" although it actually should render as this:

image

I suggest we take actual EBNF for LaTeX and implement that instead of relying on intuition :)

B3zaleel:
Added the functionality of command arguments being passed without braces. Such as "\infty" in \int_3^\infty, \frac17 and \sqrt\beta.

Closes #60.

TODO (by @ForNeVeR):

- [x] rebase on top of `master`
- [x] add some unit tests
  - [x] one test for each of hardcoded commands (`\frac` etc.)
  - [x] ~~add a test for generic commands (the ones that can be defined in XML)~~ _not applicable_
  - [x] add test for scripts like `_x^y`
  - [x] add test for text style (`\text` or `\mathrm`? See around `textStyles.Contains(command)`)
  - [x] add tests for `\sqrt`'s optional argument (e.g. `\sqrt[2]2`)
- [x] generalize the method (should work for all arguments)

### DIFF
--- a/src/WpfMath.Tests/ParserTests.fs
+++ b/src/WpfMath.Tests/ParserTests.fs
@@ -241,3 +241,19 @@ let ``\underline should parse arguments properly``(text : string) : unit =
     assertParseResult
     <| text
     <| (formula (row <| seq { yield upcast underline(char '1'); yield! ``123`` }))
+
+[<Theory>]
+[<InlineData("x^y_z");
+  InlineData("x^y_{z}");
+  InlineData("x^{y}_z");
+  InlineData("x^{y}_{z}");
+  InlineData("{x}^y_z");
+  InlineData("{x}^y_{z}");
+  InlineData("{x}^{y}_z");
+  InlineData("{x}^{y}_{z}");
+  InlineData("x^y_ z");
+  InlineData("{x} ^ {y} _ {z}")>]
+let ``Scripts should be parsed properly``(text : string) : unit =
+    assertParseResult
+    <| text
+    <| (formula <| scripts (char 'x') (char 'z') (char 'y'))
--- a/src/WpfMath.Tests/ParserTests.fs
+++ b/src/WpfMath.Tests/ParserTests.fs
@@ -233,6 +233,16 @@ let ``\sqrt should parse arguments properly``(text : string) : unit =
     <| (formula (row <| seq { yield upcast radical(char '1'); yield! ``123`` }))
 
 [<Theory>]
+[<InlineData(@"\sqrt [2]1123");
+  InlineData(@"\sqrt [ 2]{1}123");
+  InlineData(@"\sqrt[2 ] 1123");
+  InlineData(@"\sqrt[ 2 ] {1}123")>]
+let ``\sqrt should parse optional argument properly``(text : string) : unit =
+    assertParseResult
+    <| text
+    <| (formula (row <| seq { yield upcast radicalWithDegree (char '2') (char '1'); yield! ``123`` }))
+
+[<Theory>]
 [<InlineData(@"\underline1123");
   InlineData(@"\underline{1}123");
   InlineData(@"\underline 1123");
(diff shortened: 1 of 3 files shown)


alexreg:
There are some code formatting issues. Also, the exception messages could be better (and perhaps defined in one place, be it a constant or a constructor or whatever).

ForNeVeR:
That definitely requires a ton of new tests.

B3zaleel:
Not all commands can have that functionality (like the color and matrix commands) and the end of source check is casual but necessary. Reading groups should be separate from reading ungrouped items(where we just want some linearly joined items).

ForNeVeR:
Alright, I didn't knew that.

Then could you extract the repeating code to a _new_ method named `ReadGroupOrArgument`?

B3zaleel:
Now its done with this method, "GetSimpleFractionAtom".

ForNeVeR:
Wait, why have you closed this?

B3zaleel:
I deleted the branch because I thought it wasn't making any useful contribution or it wasn't doing what it was supposed to do the right way.

ForNeVeR:
It *was* doing useful contribution!

I'm not always able to respond in a timely manner, sorry for that. If I think your PR is unnecessary — I will say that. That wasn't the case here. I'm actually looking forward to review that again and merge the brach if it won't have any critical issues when I'll have some spare time.

B3zaleel:
I think now it should be up and running again, sorry for the misunderstanding.

ForNeVeR:
> Not all commands can have that functionality (like the color and matrix commands) 

I have read a bit about the LaTeX command system, and now I think that every command supports this.

With some commands, you just have no way of exploiting this mode (because they take complex arguments), and `matrix` you mentioned is not a command but an environment and thus has its' own rules.

⚠ Please, if you have examples of commands that does not use this mode, i.e. `\cmd4` is not the same as `\cmd{4}` — give me the examples, that's important!

I will now take over your branch (thanks a lot for a basic implementation, I hope to reuse that!), add some tests and then we'll review that again. Thanks!

Happypig375:
@ForNeVeR [Numbers are _NOT_ allowed as command names.](https://tex.stackexchange.com/a/28918)

ForNeVeR:
@Happypig375 I'm sorry but I can't see how it contradicts with what I say,

`\frac11` is definitely a command `frac` applied to two arguments, right? Is it right for any command instead of `frac`?

Happypig375:
My comment is highlighting that no exceptional commands as you asked can exist. I'm not contradicting what you say.

ForNeVeR:
I wasn't talking about _that_ kind of exceptions, but about another kind. Look, I'm not a LaTeX expert, so some of this may be actually absurd, and I'd like to hear if it is.

Imagine kind of command that _parses its own arguments_. Let's call it `myCommand`. So, when I write `\myCommand44`, it doesn't parse as `\myCommand{4}{4}`, but instead will be parsed according to the custom rules of `myCommand`: for example, it may postulate that it takes any number of numbers (pun not intended!). So, `\myCommand44` will be parsed as `\myCommand{44}`.

May command like that exist?

Happypig375:
Here is what I know:
- [User-defined commands _can_ override the default argument parsing by using `\def`.](https://tex.stackexchange.com/questions/81667/the-different-meanings-of-latexs-stop-command)
- Primitive commands parse differently than the default argument parsing almost all the time. For example:
  - Command declaration commands like `\def` has a wildly different syntax than the rest.
- All commands that do not override the default argument parsing can have their braces omitted. Even `\usepackage` supports this.

ForNeVeR:
@Happypig375 thanks for your useful advices.

I've added a TODO section to the first comment of this PR. I'll follow the plan.

ForNeVeR:
@B3zaleel, thanks a lot for your contribution! It was finally merged.

PR #152, opened 2018-08-22
Move converters to namespace

B3zaleel:


Happypig375:
Are you sure this is backwards-compatible?

B3zaleel:
The main code hasn't changed, just the namespace. In case we need to add converters for other mathematical writing formats, it might pollute the main directory.

ForNeVeR:
@B3zaleel yes, **it is a breaking change**. In case any library user uses the converter in the current namespace, their code will break with library update.

Although, technically, we **could** still make breaking changes (because we're still in `0.x` version range), and I welcome everybody to share their thoughts on the matter. It's better to break things right now, **before** we've entered the 1.0 version range.

I was going to deny that breaking change, but @B3zaleel made a really good argument about possibility of adding new converters.

ForNeVeR:
Thanks for your contribution!

PR #153, opened 2018-09-05
#50 Using clear type instead of normal anti-aliasing

B3zaleel:


src/WpfMath/Controls/FormulaControl.xaml.cs
```diff
@@ -111,6 +119,20 @@ public FormulaControl()
...
+            
+            if(IsCharacterSmoothingEnabled)
+            {
+                base.VisualTextRenderingMode = TextRenderingMode.ClearType;
+                base.VisualTextHintingMode = TextHintingMode.Fixed;
+                base.VisualEdgeMode = EdgeMode.Aliased;
+                base.VisualClearTypeHint = ClearTypeHint.Enabled;
+            }
+       
+        }
```
ForNeVeR: I can't see the purpose of setting these properties *after* calling `base.OnRender`.

If they affect render, then they should be set before calling `base.OnRender`. If they don't, then they should be set in the dependency property callback instead.

ForNeVeR:
@B3zaleel, thanks for your work on the matter, but I agree with @gsomix: we'll document your findings in the documentation section. Personally I like the old rendering mode, but I understand that it's not always the preferable one. The rendering properties are already controllable by user, so we don't need to make any particular improvements in WPF-Math here.

ForNeVeR:
Superseded by #194.

PR #155, opened 2018-09-06
Support for tabular atoms(matrices, multicolumn atoms, etc)

### LINKED ISSUE #100: Support for matrix elements
First off I have to say that this is extremely useful to me. However, I could not find any support for implementing matrices here are there any plans to do that? 

Also, are there any plans to write a technical documentation on how the code is organized, and how to implement new functions etc? 

Kenneth

B3zaleel:
Added an atom for rendering matrices. 
As at now matrices can be processed using a command (as in"*\\matrix{...}*", just like in the Equation Editor in Microsoft Word). Although it can be processed using the "*\\begin{...}...\end{...}*" command(which I have implemented), I've held that back because it would end up making the *TexFormulaParser* very messy/crowded.

Closes #100.

### Important changes

1. We have a new `Atom` type: `MatrixAtom`, which supports everything we need for commands like `matrix`, `pmatrix`, `cases`.
2. We have a `NullAtom` for cases when the rendering isn't necessary (as a placeholder for empty matrix cells) — _not sure if we really need this, we could live without it_.
3. `MatrixCommandParser` is a `ICommandParser` (these were introduced a couple of days ago, in #200). I like the concept and overall experience of implementing the commands as `ICommandParser`.
4. There's a new `ICommandEnvironment` concept. `ICommandEnvironment` is a special context that could define its own commands and also controls how these commands flow through the parser. Whenever we want to parse any kind of child expression (e.g. nested group or command argument or whatever), we ask the `ICommandEnvironment` to provide a fresh environment for that context. 

   I had to add this to create "temporary commands" such as `\\` and `\cr` that are only available inside of matrices, and only as the top level expressions (e.g. `\matrix{ {\cr} }` is invalid).

   `ICommandEnvironment` also gets consulted by the parser whenever it encounters an unknown character. In our case, it is used by `&` — because `&` couldn't be an ordinary command because it doesn't comply to the usual command syntax (no leading `\` required).

   While implementing an `ICommandEnvironment`, I had to return `null`s in a couple of places over the code. For now, this is only mentioned in comments of the corresponding methods, but I hope C# 8 will be released soon, and we'll mark these places with nullable reference types.
  `ICommandEnvironment` is never `null`, thus there's a simple `DefaultCommandEnvironment` implementation.
5. Approval tests now support serialization for `Box`es, which helps to test box-related stuff.

### TODO:
- [x] add tests for the `CreateBox` method
- [x] Fix the tests in the console runner (I've reproduced the CI issue)
- [x] Currently blocked on #200
- [x] Fix all the review issues in the comments below
- [x] Check if there're any TODOs remaining in the code
- [x] Make sure `SourceSpan` for matrix includes the `\matrix` command itself
- [x] Refactor the `GetMatrixData` method
- [x] Compare the results with the existing LaTeX implementations
- [x] Document the work of the matrices (e.g. matrix types, separators etc.)
- [x] Perform a separate review of the resulting code
- [x] List the architectural highlights and discuss with @gsomix


src/WpfMath/TexFormulaParser.cs
```diff
@@ -273,6 +284,55 @@ private SourceSpan ReadGroup(TexFormula formula, SourceSpan value, ref int posit
             position++;
             return value.Segment(start, position - start - 1);
         }
+        
+        private string ReadGroup(string str, char leftchar, char rightchar, int startPosition)
```
ForNeVeR: This method uses a wrong approach to markup parsing.

We usually don't copy the original string, but operate on `SourceSpan`s that all point to the original string. Probably we could strip this method and use an existing `ReadGroup` instead.

src/WpfMath/TexFormulaParser.cs
```diff
@@ -436,11 +585,209 @@ private TexFormula ReadScript(TexFormula formula, SourceSpan value, ref int posi
...
+                if (colsvalid == matrixcells.Count)
+                {
+                    return matrixcells;
+                }
+                else
+                {
+                    throw new TexParseException("The column numbers are not equal.");
+                }
+            }
+        }
```
ForNeVeR: This method is very big and should be reduced into smaller and more testable pieces. Also it seems that it's missing `&`… somewhere. Will need additional debugging.

ForNeVeR:
Also: let's maybe only make the part without an `\environment` for now.

ForNeVeR:
@B3zaleel this PR took much work from you, and I appreciate it very much. I'll take over it and drive it towards the eventual merge with the master branch. Thank you!

src/WpfMath/TexFormulaParser.cs
```diff
@@ -412,11 +561,209 @@ private static SourceSpan ReadElement(SourceSpan value, ref int position)
...
+        /// Retrives the cells of a matrix from a given source.
+        /// </summary>
+        /// <param name="matrixsource">The source to retrieve the 2D-Array of atoms from.</param>
+        private List<List<Atom>> GetMatrixData(TexFormula formula, SourceSpan matrixsource, bool allowEmptyCells = true, bool parseasArrays = false)
+        {
+            List<List<StringBuilder>> rowdata = new List<List<StringBuilder>>() { new List<StringBuilder>() { new StringBuilder() } };
+            int rows = 0;
+            int cols = 0;
+            //how many characters the next row should skip for its sourcespan to start
+            int rowindent = 0;
```
ForNeVeR: This variable most of the time holds invalid values, beause it gets assigned on the parsing stage, but gets used much later, when it has some last value from the matrix.

src/WpfMath/TexFormulaParser.cs
```diff
@@ -76,14 +77,24 @@ private static void Initialize()
 
             commands = new HashSet<string>
             {
+                "\\",
+                "amatrix",
+                "bmatrix",
+                "Bmatrix",
```
ForNeVeR: `amatrix` and `bmatrix` are environments, let's remove them for now.

src/WpfMath/TexFormulaParser.cs
```diff
@@ -76,14 +77,24 @@ private static void Initialize()
...
                 "left",
+                "matrix",
                 "overline",
+                "pmatrix",
                 "right",
                 "sqrt",
-                "underline"
+                "underline",
+                "vmatrix",
+                "Vmatrix",
```
ForNeVeR: `vmatrix` is an environment, too.

src/WpfMath/TexFormulaParser.cs
ForNeVeR: Looks like I've fixed the issue with `&`.

ForNeVeR:
The expression `\matrix{\}` currently fails, we'll need to fix it.

src/WpfMath/TexFormulaParser.cs
ForNeVeR: To fix this method, I'll need to partially implement #157 first.

src/WpfMath/Atoms/NullAtom.cs
```diff
@@ -0,0 +1,16 @@
...
+
+namespace WpfMath.Atoms
+{
+    internal class NullAtom : Atom
+    {
+        public NullAtom(SourceSpan source = null, TexAtomType type = TexAtomType.Ordinary) : base(source, type)
+        {
+        }
+
+        public static Box NullBox =>new StrutBox(0, 0, 0, 0);
```
ForNeVeR: Please drop this, it is unused.

gsomix:
@ForNeVeR I'll take a look by the end of this week.

PR #156, opened 2018-09-13
Support for external fonts

B3zaleel:
Made the formula parser to have instance constructors
Restructured the parser to have support for external symbols and text style dependent Greek letters.
Created a scheme for using different formula settings and symbols(by zipping and unzipping the packaged file).


ForNeVeR:
@B3zaleel, thank you, this is very cool. How did you generate this `.wmpkg` file? It includes very interesting XML files I'd like to know about.

B3zaleel:
I used the fonttools Python module to extract the symbols from the font file and FontForge to read some other information like the general parameters (sup1, bigopspacing1, etc) and to change the location of certain glyphs.

Issue #157, opened 2018-09-16
Refactor C#-defined command system


ForNeVeR:
For now, when defining a command in C# code, I need to add it to two places:
1. `TexFormulaParser.commands` hash set.
2. `TexFormulaParser.ProcessCommand` method.

It looks like a lot of duplication comes from this, and also this system is not user-extendable.

What I imagine is the following:

1. We need to declare one command dictionary, no name duplication should ever be used.
2. Create a `TexCommand` class and allow the implementation to deal with parsed arguments etc.
3. Add an external `RegisterCommand` API for our users to define their own commands in C#.

B3zaleel:
We could create a *Command* delegate that returns an *Atom*, store the namesof supported commands and their methods in a dictionary and invoke the command in the "ProcessCommand" method. However, if we want users to create their own commands, use the atoms we create and manipulate other aspects of the library, we have to make them go public.

ForNeVeR:
Well, yep, you're right about the public Atoms, I haven't thought about that yet. That'll be little stressful, I'll need to think more. I don't wanna to publicize all our internal stuff, because that means that any changes to that stuff will be API-breaking.

ForNeVeR:
@Polina7 let's begin with simpler commands: everything except `\left`, `\right`, and `\color*`.

PR #159, opened 2018-09-18
Update command processing system.

B3zaleel:
#157 Using delegates to process commands and to extend the commands that can be passed.


ForNeVeR:
Thanks for your contribution. I believe we've already merged a contradicting decision in PR #200. It ended up being more complicated than this prototype (and we haven't yet ported all the commands onto it).

Unfortunately, I can't see a way moving this particular PR forward. I appreciate all of your work very much, but this particular piece doesn't fit into the current project design.

Issue #158, opened 2018-09-16
\color{red} crashes parser


ForNeVeR:
Open example program, and enter `\color{red}`. In my case (I was in branch #150) the program has been crashed.

I'm not sure if that's reproducible in the `master` branch but just writing it down for now. Will check after 150 merge.

PR #161, opened 2018-09-22
Fixed issue #158

B3zaleel:


ForNeVeR:
We've found that the issue is already fixed by other commits, and added tests in #223. Thanks for your investigation though!

Issue #99, opened 2017-12-28
Foreground property not working on FormulaControl


samirem:
Setting the Foreground does not change the color of the equation. As far as I know, the only way to change the color is to wrap the string in a \color{...}, but it is not always suitable.

ForNeVeR:
Well, yes, that's a good idea to set the formula default color from the control `Foreground` property. Currently it doesn't work that way, we'll need to fix that. Thanks for the report!

Bez99:
Based on my version, you could use the fgcolor command and put in your desired hex color code or color ratio. Currently it's still being tested.

PR #162, opened 2018-09-22
#99 Support for renderer foreground and background.

B3zaleel:


src/WpfMath/Boxes/CharBox.cs
```diff
@@ -35,7 +35,7 @@ internal GlyphRun GetGlyphRun(double scale, double x, double y)
 
         public override void RenderTo(IElementRenderer renderer, double x, double y)
         {
-            var color = this.Foreground ?? Brushes.Black;
+            var color = this.Foreground ??(renderer.Foreground?? Brushes.Black);
```
ForNeVeR: Instead of adding the complexity here (by adding a public `Foreground` property), I think we could reduce it: let's remove the `Brushes.Black` altogether, and let the renderer to define their default colors as they wish.

src/WpfMath/Rendering/IElementRenderer.cs
```diff
@@ -9,6 +9,16 @@ namespace WpfMath.Rendering
...
     {
+        /// <summary>
+        /// Gets the foreground of this renderer.
+        /// </summary>
+        Brush Foreground { get; }
+        
+        /// <summary>
+        /// Gets the background of this renderer.
+        /// </summary>
+        Brush Background { get; }
```
ForNeVeR: As I said in another comment: let's remove these properties from the base interface. Implementations are free to keep them if required, though.

gnimor:
No update on this? I think this feature is really a show stopper when you want to work with the control in a MVVM project..

ForNeVeR:
This PR has been superseded by #287; the underlying issue is fixed.

Issue #139, opened 2018-07-06
\thinspace exception


AndrewXS:
Start the WpfMath.Example project.
Clear the starting formula in the textbox.
Enter "\\,"
An exception is thrown in TexPredefinedFormulaParser.cs

Bez99:
Are you sure this is an issue. I did exactly what you did and no Exceptions were thrown.

AndrewXS:
I'm running the WpfMath 0.5 code from 2018-05-08. Debug build in VS Enterprise 2017 Version 15.5.6. Just tried again, same problem. You entered the backslash and the comma?

AndrewXS:
No, not a period. However, I just tried all of the spacing commands ( \, \; \: \! ), and they cause the exception.

The attached image shows the exception with \;. (The source is {\;})

![wpfmath_exception](https://user-images.githubusercontent.com/40889744/42620117-8bf4bb5a-85a9-11e8-8d6e-1ea308d2cbae.png)

PR #166, opened 2018-10-11
Support for measurement arguments.

B3zaleel:
This functionality can allow arguments of commands like \raise -0.5em{T} and \kern-2.7 pt J to work, with or without the use of curly braces.

### DIFF
--- a/src/WpfMath/Converters/SVGConverter.cs
+++ b/src/WpfMath/Converters/SVGConverter.cs
@@ -1,4 +1,4 @@
-﻿using System;
+using System;
 using System.Collections.Generic;
 using System.Linq;
 using System.Text;
@@ -11,6 +11,8 @@ namespace WpfMath.Converters
     public class SVGConverter
     {
         private int m_nestedLevel = 0;
+
+        public Queue<Brush> GeometryBrushes = new Queue<Brush>();
 
         public string ConvertGeometry(Geometry geometry)
         {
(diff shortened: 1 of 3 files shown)


src/WpfMath/Converters/SVGConverter.cs
```diff
@@ -12,6 +12,8 @@ public class SVGConverter
     {
         private int m_nestedLevel = 0;
 
+        public Queue<Brush> GeometryBrushes = new Queue<Brush>();
```
ForNeVeR: All changes in this file belong to a separate PR #162. They shouldn't be here.

src/WpfMath/TexFormulaParser.cs
```diff
@@ -290,6 +295,65 @@ private static SourceSpan ReadElement(SourceSpan value, ref int position)
...
+                    return false;
+                }
+                
+            }
+            else
+            {
+                measurementValue = null;
+                return false;
+            }
+        }
```
ForNeVeR: I'm sorry but I can't see how these changes contribute to anything. This method is never called.

src/WpfMath/TexPredefinedFormulaParser.cs
```diff
@@ -173,8 +173,11 @@ public override void Parse(SourceSpan source, XElement element)
                 {
                     formula = new TexFormula();
                 }
-
-                this.TempFormulas.Add(name, formula);
+                if (!this.TempFormulas.ContainsKey(name))
+                {
+                    this.TempFormulas.Add(name, formula);
+                }
```
ForNeVeR: This code is supposed to fix #139 but I don't think this is the right approach: it deliberately ignores the formulas introduced by different commands and merges them into one. I don't think that's right, even if it accidentally works for simple case with `\,` → `\thinspace`.

In other words, `<CreateFormula name="f">` in `\thinspace` should have no relation to `<CreateFormula name="f">` in `\,`. Each `<Formula>` block should get its own name scope.

src/WpfMath/TexPredefinedFormulaParser.cs
```diff
@@ -194,9 +197,13 @@ public TexFormula Result
             public override void Parse(SourceSpan source, XElement element)
             {
                 var name = element.AttributeValue("name");
-                var result = this.TempFormulas[name];
-                Debug.Assert(result != null);
-                this.Result = result;
+                if (this.TempFormulas.ContainsKey(name))
```
ForNeVeR: I don't think we should have this check here either. If the user have defined a `<Return>` block referencing a nonexisting `<Formula>`, then it should be an error. Such cases shouldn't be silently ignored.

ForNeVeR:
Ah, I've found that another solution to #139 has already been provided in #214. It's sometimes hard to be up-to-date to all the stuff.

PR #170, opened 2018-12-05
WIP: Avalonia wpf combined

ahopper:
This adds  support for Avalonia.
Some of the code is from https://github.com/ForNeVeR/wpf-math/pull/98 but modified to change as little code as possible.
This initial version makes no changes to the wpf code other than moving code that can be shared to the SharedMath project.

Most of the duplication can be prevented by factoring out resource access, and typeface and brush usage.


ForNeVeR:
@ahopper looks like use some preview build of Avalonia that's only available in their MyGet channel. Is it intentional?

If so, we shouldn't care about CI status for now. At some later point, I'll fix the scripts so they work with Avalonia's MyGet properly.

But do we really need a preview Avalonia build?

ahopper:
@ForNeVeR I only used the MyGet out of force of habbit. Just tested with the 0.7 release and it is fine. There are issues with the Skia backend and the embedded fonts so it currently only works on windows/d2d. There is an Avalonia pr https://github.com/AvaloniaUI/Avalonia/pull/2162 for this which hopefully will be merged before the next release so probably not worth worrying about CI for now.

ForNeVeR:
Thanks, this code was merged into the temporary "master" Avalonia integration branch in #178.

The current plan is outlined in https://github.com/ForNeVeR/wpf-math/issues/17#issuecomment-453836924; we'll proceed to work on the Avalonia branch in a usual way (through little pull requests — I hope).

Later this year I'll hopefully merge the Avalonia stuff into the master branch.

ForNeVeR:
And, yes, @ahopper, in case it wasn't clear: I really really appreciate your work on the project and the changes you've made to port this code to a more modern version of WPF-Math.

You, guys (I mean all the WPF-Math contributors), are the best. ❤️

PR #172, opened 2018-12-26
Fixed a crash due to Atom source being null when reading source length.

sskodje:
In certain situations, the Source of scriptsAtom is null, like on {\hat T}. This causes the nullreference exception. I'm not clear on wheter it's supposed to be able to null or not, but i believe "position - initialPosition" should always be equal to the source length anyway.


ForNeVeR:
Yes, it is intended to sometimes have `null` value, that's right.

I'm thinking about including and introducing [JetBrains external annotations](https://github.com/JetBrains/ExternalAnnotations) for our code to mark what is nullable and what isn't, but at the same time, C# 8 should be released soonish and we'll be able to do the same annotations with native C# syntax.

So, for now, I'll go ahead, add test for this case and merge your PR. Thanks a lot for your contribution!

PR #173, opened 2019-01-05
Fix \lbrack and \rbrack

### LINKED ISSUE #171: Parenthesis instead of square brackets
`\lbrack` and `\rbrack` are rendered to left and right parentheses `()`, but they should be rendered to left and right square brackets `[]`.

Moreover, the parser parses a left parenthesis `(` to a `SymbolAtom` containing the name lbrack. I guess the error above arises from the confusion about what lbrack and rbrack are.

ForNeVeR:
Closes #171.

The first commit was cherry-picked from [the commit](https://github.com/Altaxo/Altaxo/commit/60f8ca7b2da9ea057ad4a129ea707f3f31a3746a) [pointed](https://github.com/ForNeVeR/wpf-math/issues/171#issuecomment-449133395) by @lellid.

@lellid, thank you!

**TODO**:
- [x] un-apply the formatting changes
- [x] investigate what's going on with the XML diff (probably the usual CRLF stuff so doesn't matter)
  - _confirmed, the changes are from CRLF to LF, so that's okay_
- [x] add a couple of tests

### DIFF
--- a/src/WpfMath/TexFormulaParser.cs
+++ b/src/WpfMath/TexFormulaParser.cs
@@ -37,9 +37,8 @@ namespace WpfMath
 
         private static readonly string[][] delimiterNames =
         {
-            new[] { "(", ")" },
             new[] { "lbrace", "rbrace" },
-            new[] { "lsqbrack", "rsqbrack" },
+            new[] { "(", ")" },
             new[] { "lbrack", "rbrack" },
             new[] { "downarrow", "downarrow" },
             new[] { "uparrow", "uparrow" },
--- a/src/WpfMath/TexFormulaParser.cs
+++ b/src/WpfMath/TexFormulaParser.cs
@@ -1,4 +1,4 @@
-﻿using System;
+using System;
 using System.Collections.Generic;
 using System.Diagnostics;
 using System.Linq;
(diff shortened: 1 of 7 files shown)


src/WpfMath/TexFormulaParser.cs
```diff
@@ -37,6 +37,7 @@ public class TexFormulaParser
 
         private static readonly string[][] delimiterNames =
         {
+            new[] { "(", ")" },
```
ForNeVeR: Hm, I'm not sure why do we need these, will investigate.

ForNeVeR:
Okay, it was interesting. It looks like the original authors were, indeed, a bit confused about the commands for parentheses (in the standard LaTeX subset there're no such commands), so they:
1. Postulated that `\lbrack` and `\lbrack` were the commands for parentheses `()`.
2. As the `\lbrack` and `\rbrack` were the commands for the square brackets `[]`, they've invented their own `\lsqbrack` and `\rsqbrack` commands.

And here's what have been done in this PR:

1. `\lsqbrack` and `\rsqbrack` are deleted for good.
2. `\lbrack` and `\rbrack` got their original meaning (i.e. `()`).
3. Some additional twiddling with `TexDelimeter` was required, because there's no need in both `Bracket` and `SquareBracket`, but it requires `Parenthesis` now.
4. While I'm at it, I've decided to fix a typo in `Delimeter`, because the compatilibity here is already broken anyway.
5. Also, we now know the purpose behind `TexFormulaParser.delimiterNames`: it is used for `TexFormulaHelper.PutDelimiterUnder` method that's triggered with reflection from our XML definitions (looks like this particular method is unused as of now). I've thrown some documentation there, but will file a refactoring issue.

ForNeVeR:
I recommend to review this PR while ignoring the whitespace changes, because CR was replaced by CRLF in a couple of files.

ForNeVeR:
Nobody got time to review this stuff, but that's not a problem: when necessary, I'm able to review my commits myself :)

Thus I've reviewed that and I'll merge it now.

PR #178, opened 2019-02-03
Avalonia+WPF combined

ForNeVeR:
This PR will rebase the work done in #170 on top of current Avalonia branch.

- Close #170.


ahopper:
This should now work with 0.8.0 Avalonia nuget release, It would be nice to get a nuget release for the avalonia version.

PR #183, opened 2019-02-23
Set DebugType to Full to prevent losing stack traces on .NET Framework

ForNeVeR:
Otherwise, the code becomes harder to debug on .NET Framework (and we for now only support .NET Framework).


ForNeVeR:
Thanks for the review.

PR #184, opened 2019-02-28
Enable textWithExpression test (#181)

### LINKED ISSUE #181: Orphaned test
This test lacks `[<Fact>]` attribute. I tried to fix it, but it doesn't even work. Need to fix it.

https://github.com/ForNeVeR/wpf-math/blob/ea242da599fa63ad811212945e84b84f0c977740/src/WpfMath.Tests/ParserTests.fs#L169-L172

ForNeVeR:
Closes #181.

Everything is okay with this test, so I've just enabled it :)


ForNeVeR:
Thanks for the review!

PR #185, opened 2019-03-12
Exclude plain text from formula parsing

### LINKED ISSUE #141: _ doesn't render inside \text
default
default

gsomix:
- [x] Implement fix
- [x] Add tests

Closes #141

### DIFF
--- a/src/WpfMath.Tests/ParserTests.fs
+++ b/src/WpfMath.Tests/ParserTests.fs
@@ -51,6 +51,10 @@ let cyrillicText() =
     verifyParseResult @"\text{абв}"
 
 [<Fact>]
+let underscoreText() =
+    verifyParseResult @"\text{_}"
+
+[<Fact>]
 let mathrm() =
     verifyParseResult @"\mathrm{sin}"
 
--- a/src/WpfMath.Tests/ParserTests.fs
+++ b/src/WpfMath.Tests/ParserTests.fs
@@ -55,6 +55,10 @@ let underscoreText() =
     verifyParseResult @"\text{_}"
 
 [<Fact>]
+let commandsInText() =
+    verifyParseResult @"\text{\alpha \beta \unknowncommand}"
+
+[<Fact>]
 let mathrm() =
     verifyParseResult @"\mathrm{sin}"
 
--- a/src/WpfMath/TexFormulaParser.cs
+++ b/src/WpfMath/TexFormulaParser.cs
@@ -247,31 +247,26 @@ namespace WpfMath
             return formula;
         }
 
-        private TexFormula ParseChars(SourceSpan value, string textStyle)
+        private static TexFormula ParseChars(SourceSpan value, string textStyle)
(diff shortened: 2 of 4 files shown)


src/WpfMath/TexFormulaParser.cs
```diff
@@ -247,6 +247,29 @@ private TexFormula Parse(SourceSpan value, ref int position, bool allowClosingDe
             return formula;
         }
 
+        private static TexFormula ParseChars(SourceSpan value, string textStyle)
```
ForNeVeR: Let's find a better name for this. `ParseRawText`? Or maybe even `ConvertRawText`, since we aren't actually parsing anything?

src/WpfMath.Tests/ParserTests.fs
```diff
@@ -50,6 +50,10 @@ let spacesInText() =
 let cyrillicText() =
     verifyParseResult @"\text{абв}"
 
+[<Fact>]
+let underscoreText() =
+    verifyParseResult @"\text{_}"
```
ForNeVeR: Let's add another interesting test here: `\text{\alpha \beta \unknowncommand}`: it will show how the algorithm handles spaces and `\commands` (it should not change them actually).

src/WpfMath/TexFormulaParser.cs
```diff
@@ -247,6 +247,29 @@ private TexFormula Parse(SourceSpan value, ref int position, bool allowClosingDe
...
+        {
+            var localPosition = 0;
+            return ParseChars(value, ref localPosition, textStyle);
+        }
+
+        private static TexFormula ParseChars(SourceSpan value, ref int position, string textStyle)
```
ForNeVeR: Having that said, I think that this overload is redundant.

We'll never want to pass `ref int position` to `ConvertRawText`, because nobody should care about the extracted position from the text; it will always convert the text completely, so `position` will always point to the `value` end in the source text.

So, let's merge these overloads and get rid of `ref int position`?

src/WpfMath/TexFormulaParser.cs
```diff
@@ -487,10 +510,21 @@ private static SourceSpan ReadElement(SourceSpan value, ref int position)
...
+                switch (command)
+                {
+                    case TexUtilities.TextStyleName:
+                        styledFormula = ParseChars(ReadElement(value, ref position), command);
+                        break;
+                    default:
+                        styledFormula = Parse(ReadElement(value, ref position), command);
+                        break;
+                }
+
```
ForNeVeR: Are you sure we want `switch` here? Looks like it won't ever be expanded to cover any other commands? I'd use ternary expression.
```suggestion
var styledFormula = command == TexUtilities.TextStyleName
    ? ParseChars(ReadElement(value, ref position), command)
    : Parse(ReadElement(value, ref position), command);
```

gsomix:
Done!

PR #199, opened 2019-08-13
Migrate to .NET 4.5.2

### LINKED ISSUE #198: Get rid of unsupported frameworks
According to the official Microsoft statements, the support for .NET Framework versions <4.5.2 has been officially ended.

[Here's a Wikipedia table that summarizes the dates](https://en.wikipedia.org/wiki/.NET_Framework_version_history#Overview).

I don't think we want to support `net40` that long. It's time we may migrate to 4.5.2 at least.

Note that dropping support of old frameworks is a breaking change and we should probably increment the major version if we're going to drop it.

ForNeVeR:
Closes #198.

This will allow me to use `IReadOnlyDictionary` where I need to (e.g. useful for #157).

Note that the unit test project have to use .NET 4.6.1 because of the test libraries we use. I wouldn't consider that as an issue though.


ForNeVeR:
Thanks for your review.

PR #200, opened 2019-08-13
Initial implementation of parser extensions

ForNeVeR:
This is a first step towards #157. It introduces an ability to pass a command parser collection into a `TexFormulaParser` constructor (this mechanism is unused yet, but will be used in #155 soon), and, as an excercise, ports one of our command parsers to the new system.

Other parsers will be ported later in #157. I don't expect anything extraordinary there, but we'll probably be able to share more parts of the parser implementations, and thus remove some code from the `TexFormulaParser`.

Tests will help to make sure there's absolutely no observable difference between the old and the new parser code.

### DIFF
--- a/src/WpfMath/Parsers/CommandParsers.cs
+++ b/src/WpfMath/Parsers/CommandParsers.cs
@@ -1,0 +1,66 @@
+using WpfMath.Atoms;
+using WpfMath.Exceptions;
+
+namespace WpfMath.Parsers
+{
+    /// <summary>A context that will be passed to the command parser.</summary>
+    internal class CommandContext
+    {
+        /// <summary>TeX formula parser that calls the command.</summary>
+        public TexFormulaParser Parser { get; }
+
+        /// <summary>Current formula.</summary>
+        public TexFormula Formula { get; }
+
+        /// <summary>Source of the current formula. Includes both command name and the arguments.</summary>
+        public SourceSpan FormulaSource { get; }
+
+        /// <summary>
+        /// A position inside of source where the command name start. Useful to provide the source information, not for
+        /// the parsing itself.
+        /// </summary>
+        public int CommandNameStartPosition { get; }
+
+        /// <summary>
+        /// A position inside of source where the command arguments start. Should be a parser start position.
+        /// </summary>
+        public int ArgumentsStartPosition { get; }
+
+        public CommandContext(TexFormulaParser parser, TexFormula formula, SourceSpan formulaSource, int commandNameStartPosition, int argumentsStartPosition)
+        {
+            Parser = parser;
+            Formula = formula;
+            FormulaSource = formulaSource;
+            CommandNameStartPosition = commandNameStartPosition;
+            ArgumentsStartPosition = argumentsStartPosition;
+        }
+    }
+
+    internal class CommandProcessingResult
+    {
+        /// <summary>A parsed atom.</summary>
+        public Atom Atom { get; }
+
+        /// <summary>
+        /// A position pointing to the part of the <see cref="CommandContext.FormulaSource"/> where the parsing should
+        /// proceed.
+        /// </summary>
+        public int NextPosition { get; }
+
+        public CommandProcessingResult(Atom atom, int nextPosition)
+        {
+            Atom = atom;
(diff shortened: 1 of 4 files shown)


src/WpfMath/Parsers/CommandParsers.cs
```diff
@@ -0,0 +1,66 @@
...
+    internal class CommandContext
+    {
+        /// <summary>TeX formula parser that calls the command.</summary>
+        public TexFormulaParser Parser { get; }
+
+        /// <summary>Current formula.</summary>
+        public TexFormula Formula { get; }
+
+        /// <summary>Source of the current formula. Includes both command name and the arguments.</summary>
+        public SourceSpan FormulaSource { get; }
```
ForNeVeR: It is not a "source of the current formula", i.e. it doesn't necessarily correlate with the `Formula` member. Let's better call it a `CommandSource` or something?

src/WpfMath/Parsers/CommandParsers.cs
```diff
@@ -0,0 +1,66 @@
...
+        public int NextPosition { get; }
+
+        public CommandProcessingResult(Atom atom, int nextPosition)
+        {
+            Atom = atom;
+            NextPosition = nextPosition;
+        }
+    }
+
+    /// <summary>A command parser interface. Inheritors of this class can perform parsing</summary>
```
ForNeVeR:
```suggestion
    /// <summary>A parser for a particular command that should read the command name and its arguments and produce an <see cref="Atom"/>, and also move the parser position towards the end.</summary>
```

src/WpfMath/Parsers/StandardCommands.cs
```diff
@@ -0,0 +1,29 @@
...
+                var source = context.FormulaSource;
+                var position = context.ArgumentsStartPosition;
+                var underlineFormula = context.Parser.Parse(
+                    TexFormulaParser.ReadElement(source, ref position),
+                    context.Formula.TextStyle);
+                var start = context.CommandNameStartPosition;
+                var atomSource = source.Segment(start, position - start);
+                var atom = new UnderlinedAtom(atomSource, underlineFormula.RootAtom);
+                return new CommandProcessingResult(atom, position);
+            }
```
ForNeVeR: In the future, I want to make a base class for these simple parsers and move the common parts into it. But this will work for now.

src/WpfMath/TexFormulaParser.cs
```diff
@@ -434,6 +439,14 @@ private static SourceSpan ReadElement(SourceSpan value, ref int position)
                     }
             }
 
+            if (_commandRegistry.TryGetValue(command, out var parser))
+            {
+                var context = new CommandContext(this, formula, value, start, position);
+                var parseResult = parser.ProcessCommand(context);
+                position = parseResult.NextPosition;
```
ForNeVeR: Let's add an assertion here: `parseResult.NextPosition > position`; otherwise the parser didn't made any progress.

Also, add a test that declares a defunct parser and checks if an assertion works.

src/WpfMath/Parsers/CommandParsers.cs
gsomix: `CommandSource` is fine.

src/WpfMath/TexFormulaParser.cs
gsomix: Double it. Another solution is introduce assertion to command parser.

ForNeVeR: I want the command parser to stay an interface. In future, it will be an external-visible API (so the users may create their own parsers), so I'd like to currently concentrate the checks on "our" side of the API. So we can't make the check in the command parser (especially in its current state when there's no common code between parsers, because we only have one parser).

gsomix: OK, let’s do that.

ForNeVeR:
There was an additional concern I've found during the testing: `TexFormulaParser.embeddedCommands` static collection doesn't take into account the new commands provided via `_commandRegistry`.

ForNeVeR:
@gsomix, thanks for your review!

PR #204, opened 2019-09-15
Separate layers for the formula background and foreground

### LINKED ISSUE #128: \colorbox renders root index invisible

ForNeVeR:
- Closes #128.
- Closes #146 (I've incorporated the useful code from that PR).


ForNeVeR:
Thanks for your review!

Issue #203, opened 2019-09-09
`\colorbox{red}{}` crashes the parser


ForNeVeR:
In the current master, `\colorbox{red}{}` will crash the parser with the following exception:
```
System.NullReferenceException: Object reference not set to an instance of an object.
   at WpfMath.Atoms.DummyAtom..ctor(TexAtomType type, Atom atom, Boolean isTextSymbol) in D:\X-Files\Projects\wpf-math\src\WpfMath\Atoms\DummyAtom.cs:line 10
   at WpfMath.Atoms.DummyAtom..ctor(Atom atom) in D:\X-Files\Projects\wpf-math\src\WpfMath\Atoms\DummyAtom.cs:line 17
   at WpfMath.Atoms.RowAtom.CreateBoxCore(TexEnvironment environment) in D:\X-Files\Projects\wpf-math\src\WpfMath\Atoms\RowAtom.cs:line 117
   at WpfMath.Atoms.Atom.CreateBox(TexEnvironment environment) in D:\X-Files\Projects\wpf-math\src\WpfMath\Atoms\Atom.cs:line 20
   at WpfMath.Atoms.StyledAtom.CreateBoxCore(TexEnvironment environment) in D:\X-Files\Projects\wpf-math\src\WpfMath\Atoms\StyledAtom.cs:line 37
   at WpfMath.Atoms.Atom.CreateBox(TexEnvironment environment) in D:\X-Files\Projects\wpf-math\src\WpfMath\Atoms\Atom.cs:line 20
   at WpfMath.TexFormula.CreateBox(TexEnvironment environment) in D:\X-Files\Projects\wpf-math\src\WpfMath\TexFormula.cs:line 118
   at WpfMath.TexFormula.GetRenderer(TexStyle style, Double scale, String systemTextFontName) in D:\X-Files\Projects\wpf-math\src\WpfMath\TexFormula.cs:line 51
   at WpfMath.Controls.FormulaControl.Render() in D:\X-Files\Projects\wpf-math\src\WpfMath\Controls\FormulaControl.xaml.cs:line 126
   at WpfMath.Controls.FormulaControl.OnRenderSettingsChanged(DependencyObject d, DependencyPropertyChangedEventArgs e) in D:\X-Files\Projects\wpf-math\src\WpfMath\Controls\FormulaControl.xaml.cs:line 186
   at System.Windows.DependencyObject.OnPropertyChanged(DependencyPropertyChangedEventArgs e)
[...]

Issue #176, opened 2019-01-21
\frac{} (with empty radicand) crashes with nullreference exception.


sskodje:
I have previously submitted at fix for this, but it seems to have been deleted in commit a6e008e1c9cb053816aed05f51793e6b159b567f .

ForNeVeR:
Wow thanks! That's what could happen if I don't cover the code with unit tests. Will be fixed soonish, after we sort out #110.

ForNeVeR:
It is already fixed in master, but I'll go ahead and write a test for it.

PR #205, opened 2019-10-01
Fix the crashing parser by empty (#203)

rstm-sf:
That's the right decision?

### DIFF
--- a/src/WpfMath/Atoms/RowAtom.cs
+++ b/src/WpfMath/Atoms/RowAtom.cs
@@ -37,9 +37,7 @@ namespace WpfMath.Atoms
         public RowAtom(SourceSpan source, IList<TexFormula> formulaList)
             : this(
                 source,
-                formulaList
-                    .Where(formula => formula.RootAtom != null)
-                    .Select(formula => formula.RootAtom))
+                formulaList.Select(formula => formula.RootAtom))
         {
         }
 
--- a/src/WpfMath/Atoms/RowAtom.cs
+++ b/src/WpfMath/Atoms/RowAtom.cs
@@ -34,13 +34,6 @@ namespace WpfMath.Atoms
             ligatureKernChangeSet.Set((int)TexAtomType.Punctuation, true);
         }
 
-        public RowAtom(SourceSpan source, IList<TexFormula> formulaList)
-            : this(
-                source,
-                formulaList.Select(formula => formula.RootAtom))
-        {
-        }
-
         public RowAtom(SourceSpan source, Atom baseAtom)
             : this(
                 source,
@@ -79,7 +72,7 @@ namespace WpfMath.Atoms
 
         public RowAtom Add(Atom atom)
         {
-            if (atom is null)
+            if (atom is null) // TODO[F]: Mark the parameter as non-nullable and drop this check whe porting to C# 8
             {
                 return new RowAtom(this.Source, this.PreviousAtom, this.Elements);
             }
--- a/src/WpfMath/Atoms/RowAtom.cs
+++ b/src/WpfMath/Atoms/RowAtom.cs
@@ -58,7 +58,8 @@ namespace WpfMath.Atoms
 
         internal RowAtom(SourceSpan source, IEnumerable<Atom> elements)
             : base(source) =>
-            this.Elements = elements.Where(x => x != null ).ToList().AsReadOnly();
+            this.Elements = elements.Where(x => x != null).ToList().AsReadOnly();
+            // TODO[F]: Fix this with C# 8 migration: there shouldn't be nullable atoms in this collection
 
         public DummyAtom PreviousAtom { get; }
 
--- a/src/WpfMath/Atoms/RowAtom.cs
+++ b/src/WpfMath/Atoms/RowAtom.cs
@@ -67,7 +67,7 @@ namespace WpfMath.Atoms
 
         internal RowAtom(SourceSpan source, IEnumerable<Atom> elements)
             : base(source) =>
-            this.Elements = elements.ToList().AsReadOnly();
+            this.Elements = elements.Where(x => x != null ).ToList().AsReadOnly();
 
         public DummyAtom PreviousAtom { get; }
 
@@ -81,6 +81,11 @@ namespace WpfMath.Atoms
 
         public RowAtom Add(Atom atom)
         {
+            if (atom is null)
+            {
+                return new RowAtom(this.Source, this.PreviousAtom, this.Elements);
+            }
+
             var newElements = this.Elements.ToList();
             newElements.Add(atom);
             return new RowAtom(this.Source, this.PreviousAtom, newElements.AsReadOnly());
--- a/src/WpfMath.Tests/BoxTests.fs
+++ b/src/WpfMath.Tests/BoxTests.fs
@@ -111,3 +111,7 @@ let nestedMatrixBox() =
 [<Fact>]
 let wideItemInMatrixBox() =
     verifyBox @"x = \pmatrix{0 & -r & 0 \\ 0 & 0 & -r sin^2(\theta)}"
+
+[<Fact>]
+let emptyColorbox(): unit =
+    verifyBox @"\colorbox{red}{}"
--- a/src/WpfMath.Tests/TestResults/BoxTests.emptyColorbox.approved.txt
+++ b/src/WpfMath.Tests/TestResults/BoxTests.emptyColorbox.approved.txt
@@ -1,0 +1,18 @@
+﻿{
+  "Children": [],
+  "Source": {
+    "Start": 1,
+    "End": 16,
+    "Length": 15,
+    "Source": "\\colorbox{red}{}"
+  },
+  "Foreground": null,
+  "Background": "#FFED1B23",
+  "TotalHeight": 0.0,
+  "TotalWidth": 0.0,
+  "Italic": 0.0,
+  "Width": 0.0,
+  "Height": 0.0,
+  "Depth": 0.0,
+  "Shift": 0.0
+}
--- a/src/WpfMath/Boxes/StrutBox.cs
+++ b/src/WpfMath/Boxes/StrutBox.cs
@@ -5,12 +5,7 @@ namespace WpfMath.Boxes
     // Box representing whitespace.
     internal class StrutBox : Box
     {
-        private static readonly StrutBox emptyStrutBox = new StrutBox(0, 0, 0, 0);
-
-        public static StrutBox Empty
-        {
-            get { return emptyStrutBox; }
-        }
+        public static StrutBox Empty { get; } = new StrutBox(0, 0, 0, 0);
 
         public StrutBox(double width, double height, double depth, double shift)
         {
--- a/src/WpfMath/TexFormula.cs
+++ b/src/WpfMath/TexFormula.cs
@@ -10,27 +10,6 @@ namespace WpfMath
     // Represents mathematical formula that can be rendered.
     public sealed class TexFormula
     {
-        public TexFormula(IList<TexFormula> formulaList)
-        {
-            Debug.Assert(formulaList != null);
-
-            if (formulaList.Count == 1)
-                Add(formulaList[0]);
-            else
-                this.RootAtom = new RowAtom(null, formulaList);
-        }
-
-        public TexFormula(TexFormula formula)
-        {
-            Debug.Assert(formula != null);
-
-            Add(formula);
-        }
-
-        public TexFormula()
-        {
-        }
-
         public string TextStyle
         {
             get;


rstm-sf:
Ok, I think I have to install F#.

ForNeVeR:
Hello. Could you please clarify, what problem are you trying to solve?

According to the current WPF-Math design, `formula.RootAtom` is nullable and should be treated as nullable everywhere. If you know a place where it isn't, then it's better to fix that place instead of plugging it with some non-null instances here and there.

rstm-sf:
Hello!

> According to the current WPF-Math design, `formula.RootAtom` is nullable and should be treated as nullable everywhere

I will take into account.



> Could you please clarify, what problem are you trying to solve?

Issue #203 

`remaining.RootAtom` After `remainingString` [parsed] 

After [parsing]((https://github.com/ForNeVeR/wpf-math/blob/98b46892464e3c6380dfb0127157380d91dcd286/src/WpfMath/TexFormulaParser.cs#L476)) the `remainingString`, the `remaining.RootAtom` (line 480) is null in case `\colorbox{red}{}` (I think the same case with `\color{red}{}`). It as a result leads to exception ([`atom`](https://github.com/ForNeVeR/wpf-math/blob/98b46892464e3c6380dfb0127157380d91dcd286/src/WpfMath/Atoms/DummyAtom.cs#L10) is null). I thought `NullAtom` was created for this.

rstm-sf:
> WpfMath.Tests.ParserExceptionTests.\mathrm{} should throw exn

@ForNeVeR, why do you think that an exception should occur in the case of `\mathrm{}`?

rstm-sf:
In LaTeX this normal

rstm-sf:
@ForNeVeR, should the following be [here](https://github.com/ForNeVeR/wpf-math/blob/98b46892464e3c6380dfb0127157380d91dcd286/src/WpfMath/TexFormulaParser.cs#L480)?
```c#
return new StyledAtom(source, remaining.RootAtom ?? new NullAtom(), new SolidColorBrush(color), null);
```

ForNeVeR:
@rstm-sf, thanks for clarification! Now to your questions:

> After [parsing](https://github.com/ForNeVeR/wpf-math/blob/98b46892464e3c6380dfb0127157380d91dcd286/src/WpfMath/TexFormulaParser.cs#L476) the `remainingString`, the `remaining.RootAtom` (line 480) is null in case `\colorbox{red}{}` (I think the same case with `\color{red}{}`). It as a result leads to exception ([`atom`](https://github.com/ForNeVeR/wpf-math/blob/98b46892464e3c6380dfb0127157380d91dcd286/src/WpfMath/Atoms/DummyAtom.cs#L10) is null). I thought `NullAtom` was created for this.

Thanks, now I get what the error is. I don't think that `NullAtom` was created for this, but could/should it be used for this? Well, maybe? What alternatives do we have?

We could write a code like that instead of [TexFormulaParser.cs:480](https://github.com/ForNeVeR/wpf-math/blob/98b46892464e3c6380dfb0127157380d91dcd286/src/WpfMath/TexFormulaParser.cs#L480):

```csharp
return remaining.RootAtom == null ? null : new StyledAtom(source, remaining.RootAtom, new SolidColorBrush(color), null);
```

But that creates another place in our code that could return `null` instead of a proper atom, and I think it's worse than `new StyledAtom(source, remaining.RootAtom ?? new NullAtom(), new SolidColorBrush(color), null)`.

> @ForNeVeR, why do you think that an exception should occur in the case of `\mathrm{}`?

That's a good question, actually. I think the code was written that way originally, and my tests just asserted the existing behavior. I agree that we should fix the current behavior to be more compliant with the standard TeX engines.

Care to create a separate PR for that, if you're interested?

> @ForNeVeR, should the following be [here](https://github.com/ForNeVeR/wpf-math/blob/98b46892464e3c6380dfb0127157380d91dcd286/src/WpfMath/TexFormulaParser.cs#L480)?
> ```c#
> return new StyledAtom(source, remaining.RootAtom ?? new NullAtom(), new SolidColorBrush(color), null);
> ```

Accidentally I've already answered this question earlier. Yes, please!

ForNeVeR:
Ah, yes, one more thing: if you want any help with tests and is unfamiliar with F# and/or approval tests library we use, then don't hesitate to ask. I'm ready to fix them myself.

rstm-sf:
> I don't think that `NullAtom` was created for this, but could/should it be used for this? Well, maybe? What alternatives do we have?

Maybe write some EmptyAtom? But you need to he was really is for all 'empty'.

Actually there needs to be something like that ```new StrutBox(1, 1, 0, 0)```

> Care to create a separate PR for that, if you're interested?

Thanks! But I still don’t even understand why ```remaining.RootAtom``` needs to be null. I just walked a debugger and did what I could :)

I’d better choose a simple way and I will decide _Issues_ as far as possible ;)

rstm-sf:
> Ah, yes, one more thing: if you want any help with tests and is unfamiliar with F# and/or approval tests library we use, then don't hesitate to ask. I'm ready to fix them myself.

Thanks! Yes, I am far from functional programming.

ForNeVeR:
> Maybe write some EmptyAtom? But you need to he was really is for all 'empty'.

Well, `NullAtom` seems empty enough for our purposes :)

> Actually there needs to be something like that `new StrutBox(1, 1, 0, 0)`

Why do you think we need that? I honestly don't see a purpose for this particular size. It seems to me that zero size should be okay, but any other size would need to be properly scaled etc. etc., I don't think it's of any use currently.

ForNeVeR:
I'll review the PR and fix tests in the meantime.

rstm-sf:
> Why do you think we need that?



<details>
  <summary>Because I get the following result  and did not think about the size</summary>

  
</details>

> I honestly don't see a purpose for this particular size. It seems to me that zero size should be okay, but any other size would need to be properly scaled etc. etc., I don't think it's of any use currently.

Ok, np :)

ForNeVeR:
Hm. I'm not sure if we really want to follow this pattern from the image. Let's make it empty for now.

If any user will complain about the empty block rendering as an empty block, we may reconsider :)

rstm-sf:
I looked at open issues and find similary: #158, #176. 

But now in master the #176 problem is fixed (see [FractionAtom.cs#L114](https://github.com/ForNeVeR/wpf-math/blob/master/src/WpfMath/Atoms/FractionAtom.cs#L114) and L116).

rstm-sf:
Oh, no. I was wrong. It was enough to look at the diff of commit [a6e008e](https://github.com/ForNeVeR/wpf-math/commit/a6e008e1c9cb053816aed05f51793e6b159b567f) specified by @sskodje and the current master.

rstm-sf:
Hello!

I’m thinking now and it seems that it’s better to do the same check
> [FractionAtom.cs#L114](https://github.com/ForNeVeR/wpf-math/blob/master/src/WpfMath/Atoms/FractionAtom.cs#L114) and L116

Or still stay on option with ```?? new NullAtom()``` ?

ForNeVeR:
Hm. Maybe it would be better to handle `null` parameter in the atom constructors? I.e. make the `StyledAtom` to accept `null` as the argument? What do you think?

rstm-sf:
In addition to `CreateBox` operations, there are other operations where `this.Atom` should not be null. Therefore, every time you need to write a check for null. In my opinion `NullAtom` (or other?)  is suitable for this. But I still do not understand the architecture of the parser to understand where I am mistaken.

ForNeVeR:
Okay, let's look at the situation from another side.

What if we make `RootAtom` non-nullable? We probably can assign `new NullAtom()` there in the formula constructor, and handle it everywhere else where we check for `null`.

gsomix:
Please, call it `EmptyAtom`.

ForNeVeR:
I have to say that this PR currently has many formatting changes that make it very hard to get what's been really changed. Please, try to not introduce unnecessary formatting changes in PRs not completely devoted to that purpose. It's very hard to review them.

rstm-sf:
Ok. I can delete all changes or transfer to _refactor_. What is better to do?

ForNeVeR:
I would reset this branch to `master` and then start making the changes again (i.e. rename the class and then do `RootAtom = new NullAtom()` in the formula constructor).

ForNeVeR:
Likely, C# 8 transition would help us here, but now it's blocked on your changes in this PR, because I don't want (neighter myself or you) to resolve the conflicts afterwards :)

rstm-sf:
> `RootAtom = new NullAtom()`

Actually, I’m thinking now whether it is possible to get rid of the ctor `RowAtom(SourceSpan)` ([RowAtom.cs#L55](https://github.com/ForNeVeR/wpf-math/blob/master/src/WpfMath/Atoms/RowAtom.cs#L55)) and change the ctor `RowAtom(SourceSpan, Atom)` ([RowAtom.cs#L46)](https://github.com/ForNeVeR/wpf-math/blob/0d788e639d25e24509756809cda193f14e1f4724/src/WpfMath/Atoms/RowAtom.cs#L46)) or the ctor `RowAtom(SourceSpan, IEnumerable<Atom>)` ([RowAtom.cs#L68](https://github.com/ForNeVeR/wpf-math/blob/0d788e639d25e24509756809cda193f14e1f4724/src/WpfMath/Atoms/RowAtom.cs#L68)) so that it can be `null` replaced with `new NullAtom()`.

ForNeVeR:
Is it `RowAtom` that creates a problem for us? I thought it was `TexFormula.RootAtom` that's null and this is a problem. How is `RowAtom` involved?

rstm-sf:
We should not do `RootAtom = new NullAtom()`. Must either leave `RootAtom = null` or create `NullAtom` and `EmptyAtom`.

rstm-sf:
> Is it `RowAtom` that creates a problem for us? I thought it was `TexFormula.RootAtom` that's null and this is a problem. How is `RowAtom` involved?

Sorry for delay, I had to understand for myself what I understood.

Partially the answer is given in #207 , starting with the words _Afterword_. If this is not enough, then I can try to reveal the answer in more detail.

rstm-sf:
> Actually, I’m thinking now whether it is possible to get rid of the ctor `RowAtom(SourceSpan)` ([RowAtom.cs#L55](https://github.com/ForNeVeR/wpf-math/blob/master/src/WpfMath/Atoms/RowAtom.cs#L55)) and change the ctor `RowAtom(SourceSpan, Atom)` ([RowAtom.cs#L46)](https://github.com/ForNeVeR/wpf-math/blob/0d788e639d25e24509756809cda193f14e1f4724/src/WpfMath/Atoms/RowAtom.cs#L46)) or the ctor `RowAtom(SourceSpan, IEnumerable<Atom>)` ([RowAtom.cs#L68](https://github.com/ForNeVeR/wpf-math/blob/0d788e639d25e24509756809cda193f14e1f4724/src/WpfMath/Atoms/RowAtom.cs#L68)) so that it can be `null` replaced with `new NullAtom()`.

For example, [this (PhantomAtom.cs#L20)](https://github.com/ForNeVeR/wpf-math/blob/b05b6acee7d0d6fa476469b4b2a41a2c5bb8d4f1/src/WpfMath/Atoms/PhantomAtom.cs#L20) code can be simplified to 
```c#
this.RowAtom = new RowAtom(null, baseAtom); // or maybe right new RowAtom(source, baseAtom);
```

And by the way, in [StyledAtom.cs#L12](https://github.com/ForNeVeR/wpf-math/blob/98b46892464e3c6380dfb0127157380d91dcd286/src/WpfMath/Atoms/StyledAtom.cs#L12) this is what happens

ForNeVeR:
I'll be honest: I am completely lost in your thoughts. Do we still discussing the initial problem? Should I review / add tests for that already, or not?

rstm-sf:
Sorry to start thinking about another issue.

Just start over and look at the current changes. They are designed to solve problems with #158 and #203.

ForNeVeR:
Alright, thanks for clarifying!

rstm-sf:
> > @ForNeVeR, why do you think that an exception should occur in the case of `\mathrm{}`?
> 
> That's a good question, actually. I think the code was written that way originally, and my tests just asserted the existing behavior. I agree that we should fix the current behavior to be more compliant with the standard TeX engines.
> 
> Care to create a separate PR for that, if you're interested?

Yes, it’s better to create separate PRs. But only after the adoption of these changes.

src/WpfMath/Atoms/RowAtom.cs
```diff
@@ -81,6 +79,11 @@ internal RowAtom(SourceSpan source, IEnumerable<Atom> elements)
 
         public RowAtom Add(Atom atom)
         {
+            if (atom is null)
+            {
+                return new RowAtom(this.Source, this.PreviousAtom, this.Elements);
```
rstm-sf: I do not like that I had to write the following

rstm-sf:
Hm, #161 also solved the #158 problem

src/WpfMath/Atoms/RowAtom.cs
```diff
@@ -37,9 +37,7 @@ static RowAtom()
         public RowAtom(SourceSpan source, IList<TexFormula> formulaList)
             : this(
                 source,
-                formulaList
-                    .Where(formula => formula.RootAtom != null)
-                    .Select(formula => formula.RootAtom))
+                formulaList.Select(formula => formula.RootAtom))
```
ForNeVeR: I have a better solution for this code: I'll delete it for good. The only place where it's called is an unused constructor of `TexFormula`.

src/WpfMath/Atoms/RowAtom.cs
```diff
@@ -67,7 +65,7 @@ private RowAtom(SourceSpan source, DummyAtom previousAtom, ReadOnlyCollection<At
 
         internal RowAtom(SourceSpan source, IEnumerable<Atom> elements)
             : base(source) =>
-            this.Elements = elements.ToList().AsReadOnly();
+            this.Elements = elements.Where(x => x != null ).ToList().AsReadOnly();
```
ForNeVeR: This shouldn't be necessary because `RowAtom` doesn't have any means to add `null` to its element list, and thus we shouldn't check for `null` here. The `elements` is really an enumerable of non-null atoms.

src/WpfMath/Atoms/RowAtom.cs
ForNeVeR: I understand that. I'll mark this as a code that will be improved when porting to C# 8: we should really mark this all as non-nullable, but I can see that it solves the problem at the moment.

src/WpfMath/Atoms/RowAtom.cs
ForNeVeR: I retract this statement: it looks necessary. We'll fix it with C# 8, then :)

PR #212, opened 2019-10-13
`\mathrm{}` shouldn't throw exn

rstm-sf:
Change state "\mathrm{} should throw exn" to "\mathrm{} shouldn't throw exn"

### DIFF
--- a/src/WpfMath/TexFormula.cs
+++ b/src/WpfMath/TexFormula.cs
@@ -33,7 +33,6 @@ namespace WpfMath
         public void Add(TexFormula formula, SourceSpan source = null)
         {
             Debug.Assert(formula != null);
-            Debug.Assert(formula.RootAtom != null);
 
             this.Add(
                 formula.RootAtom is RowAtom
@@ -50,7 +49,6 @@ namespace WpfMath
         /// <param name="rowSource">The source that will be set for the resulting row atom.</param>
         internal void Add(Atom atom, SourceSpan rowSource)
         {
-            Debug.Assert(atom != null);
             if (this.RootAtom == null)
             {
                 this.RootAtom = atom;
@@ -91,10 +89,7 @@ namespace WpfMath
 
         internal Box CreateBox(TexEnvironment environment)
         {
-            if (this.RootAtom == null)
-                return StrutBox.Empty;
-            else
-                return this.RootAtom.CreateBox(environment);
+            return this.RootAtom?.CreateBox(environment) ?? StrutBox.Empty;
         }
 
         internal static SystemFont GetSystemFont(string fontName, double size)
--- a/src/WpfMath/TexFormula.cs
+++ b/src/WpfMath/TexFormula.cs
@@ -1,4 +1,4 @@
-﻿using System.Collections.Generic;
+using System.Collections.Generic;
 using System.Diagnostics;
 using System.Linq;
 using System.Windows.Media;
@@ -33,6 +33,7 @@ namespace WpfMath
         public void Add(TexFormula formula, SourceSpan source = null)
         {
             Debug.Assert(formula != null);
+            Debug.Assert(formula.RootAtom != null);
 
             this.Add(
                 formula.RootAtom is RowAtom
--- a/src/WpfMath/TexFormula.cs
+++ b/src/WpfMath/TexFormula.cs
@@ -89,7 +89,10 @@ namespace WpfMath
 
         internal Box CreateBox(TexEnvironment environment)
         {
-            return this.RootAtom?.CreateBox(environment) ?? StrutBox.Empty;
+            if (this.RootAtom == null)
+                return StrutBox.Empty;
+            else
+                return this.RootAtom.CreateBox(environment);
         }
 
         internal static SystemFont GetSystemFont(string fontName, double size)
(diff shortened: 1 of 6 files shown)


rstm-sf:
I tried adding a test (like for `\colorbox{red}{}`) and getting a test fail
```f#
[<Fact>]
let emptyMathrm(): unit =
    verifyBox @"\mathrm{}"
```

Also tried to write the following for TexFormula.Add(Atom, SourceSpan)
```c#
if (this.RootAtom == null)
{
    this.RootAtom = new RowAtom(rowSource, new []{ atom });
}
```
But it didn’t help either.

It seems that I did not understand the parser.

rstm-sf:
Ah, I also had to add `BoxTests.emptyMathrm.approved.txt`

rstm-sf:
@ForNeVeR, hello!

Do you think this is a good change?

Tests fail because there is more nesting.

src/WpfMath/TexFormulaParser.cs
```diff
@@ -572,16 +572,15 @@ internal static SourceSpan ReadElement(SourceSpan value, ref int position)
...
-                var atom = AttachScripts(formula, value, ref position, styledFormula.RootAtom, true, environment);
-                var source = new SourceSpan(formulaSource.Source, formulaSource.Start, position);
-                formula.Add(atom, source);
+                var remainingString = ReadElement(value, ref position);
+                var remaining = command == TexUtilities.TextStyleName
+                    ? ConvertRawText(remainingString, command)
+                    : Parse(remainingString, command, environment.CreateChildEnvironment());
+
+                var source = value.Segment(start, position - start);
+                var styledAtom = new StyledAtom(source, remaining.RootAtom, null, null);
```
ForNeVeR: I don't understand why we need this one. It is, um, a styled atom with no style added.

rstm-sf: Because it works! :)

What served to make this decision:
- the message in `throw new TexParseException("Styled text can't be empty!");`
- case similarity with `colorbox`

If I don’t add a wrapper, I will end up with a drop in the test [here (BoxTests.fs#L96)](https://github.com/ForNeVeR/wpf-math/blob/7944747d6219f77b3f0f5dcd3a723d1c92a94dd5/src/WpfMath.Tests/BoxTests.fs#L96).

ForNeVeR: I understand why you've removed the `throw` statement, and that's good!

Why does that test break though?

rstm-sf: Because after `parse source` comes `null`

rstm-sf:
Well, maybe it should be so.

src/WpfMath/TexFormulaParser.cs
ForNeVeR: Could you please clarify? I don't understand. It is here, in the local variable. How could it turn `null`?

rstm-sf: Because `TexFormula.RootAtom` is nullable. The situation is the same as, for example, with `colorbox` ([TexFormulaParser.cs#L480](https://github.com/ForNeVeR/wpf-math/blob/7944747d6219f77b3f0f5dcd3a723d1c92a94dd5/src/WpfMath/TexFormulaParser.cs#L480), [TexFormulaParser.cs#L591](https://github.com/ForNeVeR/wpf-math/blob/7944747d6219f77b3f0f5dcd3a723d1c92a94dd5/src/WpfMath/TexFormulaParser.cs#L591)).

rstm-sf: When the parsing of the formula in parentheses begins, the loops ([TexFormulaParser.cs#L217](https://github.com/ForNeVeR/wpf-math/blob/7944747d6219f77b3f0f5dcd3a723d1c92a94dd5/src/WpfMath/TexFormulaParser.cs#L217), [TexFormulaParser.cs#L293](https://github.com/ForNeVeR/wpf-math/blob/7944747d6219f77b3f0f5dcd3a723d1c92a94dd5/src/WpfMath/TexFormulaParser.cs#L293)) are skipped and `texFormula.RootAtom` remains set to `null`

rstm-sf: And, when we pass `atom.RootAtom` to the `Atom AttachScripts(...)` ([TexFormulaParser.cs#L627](https://github.com/ForNeVeR/wpf-math/blob/7944747d6219f77b3f0f5dcd3a723d1c92a94dd5/src/WpfMath/TexFormulaParser.cs#L627)), the result is `atom.RootAtom` ([TexFormulaParser.cs#L638](https://github.com/ForNeVeR/wpf-math/blob/7944747d6219f77b3f0f5dcd3a723d1c92a94dd5/src/WpfMath/TexFormulaParser.cs#L638)), i.e. `null`.

rstm-sf: And, sorry, but, I do not know what the `Atom AttachScripts(...)` should do, since I did not consider this further.

src/WpfMath/TexFormula.cs
```diff
@@ -33,7 +33,6 @@ public TexRenderer GetRenderer(TexStyle style, double scale, string systemTextFo
         public void Add(TexFormula formula, SourceSpan source = null)
         {
             Debug.Assert(formula != null);
-            Debug.Assert(formula.RootAtom != null);
```
ForNeVeR: I don't think you need to remove these assertions now (when the new code produces non-nullable items for these formulas), do you?

ForNeVeR:
Aside from my last comment, everything else looks good.

src/WpfMath/TexFormula.cs
rstm-sf: Ok

PR #213, opened 2019-10-14
Add a test for #176

### LINKED ISSUE #176: \frac{} (with empty radicand) crashes with nullreference exception.

ForNeVeR:
Closes #176 (which is actually already fixed but we have no tests for that).


ForNeVeR:
Thanks for review!

PR #218, opened 2019-10-19
Delete FSharp.Core from the project file (#210)

### LINKED ISSUE #210: Test compilation troubles in Visual Studio
While testing the project build with @NikolaiSviridov, we've discovered that our tests don't compile in Visual Studio with weird error messages.

Investigation is required.

ForNeVeR:
I've found that the FSharp.Core version is for some reason hardcoded in our current `.fsproj`, and it causes compilation troubles in VS.

FSharp.Core is implicitly added by SDK anyways, so I see no reason to keep this reference inplace.

Closes #210.


ForNeVeR:
Thanks for review!

PR #227, opened 2019-12-06
Close #151: Fix predefined function selection

### LINKED ISSUE #151: SourceSpan interference
Currently if you try to select a predefined function like sin (\sin) in the example project's Textbox, it doesn't get selected. It can only get selected when you place at least 8 characters before it (which is equal to the number of characters of the functions' actual leading characters, which in this case is "\mathrm{\sin}".  After that you have to select the characters that precede the function. Try this after clearing the textbox's content "TRYTHISFxn\cos^2(\theta) + \sin^2(\theta)=1".
Here the cos and sin seem to be bound together.

dedok1997:
Closes #151.

### DIFF
--- a/src/WpfMath/Atoms/Atom.cs
+++ b/src/WpfMath/Atoms/Atom.cs
@@ -1,19 +1,26 @@
+using System.Collections.Generic;
 using WpfMath.Boxes;
-
 namespace WpfMath.Atoms
 {
     // Atom (smallest unit) of TexFormula.
     internal abstract class Atom
     {
-        protected Atom(SourceSpan source, TexAtomType type = TexAtomType.Ordinary)
+        protected Atom(SourceSpan source, IEnumerable<Atom> children, TexAtomType type = TexAtomType.Ordinary)
         {
             this.Source = source;
             this.Type = type;
+            this.Children = children;
+        }
+
+        protected Atom(SourceSpan source, TexAtomType type = TexAtomType.Ordinary): this(source, new List<Atom>(), type)
+        {
         }
 
         public TexAtomType Type { get; }
 
-        public SourceSpan Source { get; }
+        public SourceSpan Source { get; set; }
+
+        public IEnumerable<Atom> Children { get; protected set; }
 
         public Box CreateBox(TexEnvironment environment)
         {
--- a/src/WpfMath/Atoms/RowAtom.cs
+++ b/src/WpfMath/Atoms/RowAtom.cs
@@ -46,24 +46,26 @@ namespace WpfMath.Atoms
         public RowAtom(SourceSpan source)
             : base(source)
         {
-            this.Elements = new List<Atom>().AsReadOnly();
+            this.Children = new List<Atom>().AsReadOnly();
         }
 
         private RowAtom(SourceSpan source, DummyAtom previousAtom, ReadOnlyCollection<Atom> elements)
             : base(source)
         {
             this.PreviousAtom = previousAtom;
-            this.Elements = elements;
+            this.Children = elements;
         }
 
         internal RowAtom(SourceSpan source, IEnumerable<Atom> elements)
-            : base(source) =>
-            this.Elements = elements.Where(x => x != null).ToList().AsReadOnly();
+            : base(source)
+        {
+            this.Children = elements.Where(x => x != null).ToList().AsReadOnly(); ;
+        }
             // TODO[F]: Fix this with C# 8 migration: there shouldn't be nullable atoms in this collection
 
         public DummyAtom PreviousAtom { get; }
 
-        public ReadOnlyCollection<Atom> Elements { get; }
+        public ReadOnlyCollection<Atom> Elements => (ReadOnlyCollection<Atom>)this.Children;
 
         public Atom WithPreviousAtom(DummyAtom previousAtom) =>
             new RowAtom(this.Source, previousAtom, this.Elements);
--- a/src/WpfMath/TexFormulaHelper.cs
+++ b/src/WpfMath/TexFormulaHelper.cs
@@ -153,9 +153,27 @@ namespace WpfMath
                 formulaParser.Parse(upperLimitFormula), useVerticalLimits);
         }
 
+        public int MoveOffset(Atom atom, int pos)
+        {
+            atom.Source = new SourceSpan(atom.Source.Source, pos, atom.Source.Length);
+            int currentPos = pos;
+            foreach (var child in atom.Children)
+            {
+                currentPos = MoveOffset(child, currentPos);
+            }
+            return pos + atom.Source.Length;
+        }
+
         public void AddOperator(string operatorFormula, bool useVerticalLimits)
         {
-            AddOperator(formulaParser.Parse(operatorFormula), null, null, useVerticalLimits);
+            var formula = formulaParser.Parse(operatorFormula);
+            formula.RootAtom.Source = new SourceSpan(source.Source, source.Start, formula.RootAtom.Source.Length);
+            var pos = source.Start+1;
+            foreach(var e in formula.RootAtom.Children)
+            {
+                pos = MoveOffset(e, pos);
+            }
+            AddOperator(formula, null, null, useVerticalLimits);
         }
 
         public void AddOperator(TexFormula operatorFormula, TexFormula lowerLimitFormula, TexFormula upperLimitFormula)
@@ -174,7 +192,7 @@ namespace WpfMath
             TexFormula upperLimitFormula,
             bool useVerticalLimits)
         {
-            this.Add(
+                this.Add(
                 new BigOperatorAtom(
                     operatorFormula?.RootAtom?.Source,
                     operatorFormula?.RootAtom,
--- a/src/WpfMath/TexFormulaHelper.cs
+++ b/src/WpfMath/TexFormulaHelper.cs
@@ -155,23 +155,30 @@ namespace WpfMath
 
         public int MoveOffset(Atom atom, int pos)
         {
-            atom.Source = new SourceSpan(atom.Source.Source, pos, atom.Source.Length);
-            int currentPos = pos;
-            foreach (var child in atom.Children)
+            if (atom.Source != null)
             {
-                currentPos = MoveOffset(child, currentPos);
+                atom.Source = new SourceSpan(atom.Source.Source, pos, atom.Source.Length);
+                int currentPos = pos;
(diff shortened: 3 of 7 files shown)


src/WpfMath/Atoms/Atom.cs
```diff
@@ -1,19 +1,26 @@
...
+        protected Atom(SourceSpan source, TexAtomType type = TexAtomType.Ordinary): this(source, new List<Atom>(), type)
+        {
         }
 
         public TexAtomType Type { get; }
 
-        public SourceSpan Source { get; }
+        public SourceSpan Source { get; set; }
+
+        public IEnumerable<Atom> Children { get; protected set; }
```
rstm-sf: Hello!

Excuse me.

But can we do without `IEnumerable` in public?

dedok1997: Hello. 

I guess in the current implementation it isn't possible.

But we can something like that:

1. define an abstract method in Atom
`public abstract void MoveOffset(int position);`
2. Implement that in each child of Atom

rstm-sf: we have in stock  `protected` or `internal` (based on the situation) ;)

I have another questions now. Do you want to use the interface? We can get by with interfaces like `IReadOnlyList` or `IReadOnlyCollection` instead of `IEnumerable`?

dedok1997: Yep, internal is fine.

I can use any of them. What do you recommend?

rstm-sf: > I can use any of them.

Fine :)

> What do you recommend?

less diff :)

There are no differences between your changes. It's just that one is the heir to the other and provides more options or restrictions, depending on which side you look. Just wanted to point out that using `IEnumerable` is not always good

src/WpfMath/Atoms/RowAtom.cs
```diff
@@ -46,24 +46,26 @@ public RowAtom(SourceSpan source, Atom baseAtom)
...
+            : base(source)
+        {
+            this.Children = elements.Where(x => x != null).ToList().AsReadOnly(); ;
+        }
             // TODO[F]: Fix this with C# 8 migration: there shouldn't be nullable atoms in this collection
 
         public DummyAtom PreviousAtom { get; }
 
-        public ReadOnlyCollection<Atom> Elements { get; }
+        public ReadOnlyCollection<Atom> Elements => (ReadOnlyCollection<Atom>)this.Children;
```
rstm-sf: Since you are using a `protected set`, hypothetically, the likelihood of a subsequent exception is even greater. So I would think about using a certain type

rstm-sf: Or am I wrong here, sorry

rstm-sf: Yes, this [example](https://dotnetfiddle.net/ANWesf) throws `System.InvalidCastException`

dedok1997: Sorry, my fault

rstm-sf: >this example

Sorry, something I began to complicate simple things yesterday

> Sorry, my fault

No problems. Live and learn!  :)

rstm-sf:
I tried to merge your changes with mine and I got `NRE` to `Source` in test. I think that because in the current implementation, in some cases, this is supposed to be `null`, see https://github.com/ForNeVeR/wpf-math/pull/214/files#diff-bd44197e7add12226b5df7dc12f07e0bR3

rstm-sf:
> in some cases

In `src/WpfMath/TexFormulaHelper.cs` there are methods due to which `Source` is lost. But it seems to me using `this.source` will not be the right solution

src/WpfMath/Atoms/Atom.cs
ForNeVeR: Our `Atom`s are already `internal`, so I can't see a particular point in changing the current visibility.

Wa have plans to potentially make them `public` later, and then we'll review all the available pubic API.

Also, I believe that `IEnumerable` is better here. Constructing and maintaining a `IReadOnlyList` would be more memory consuming IMO.

rstm-sf: Ok, if you think that for a property, in this situation, it is better to use `IEnumerable`, then so be it

ForNeVeR: Yes, I think so. I would be glad to hear arguments why you think that's not the best idea though. We could easily reconsider if that's harmful in some way.

rstm-sf: I apologize, but my knowledge is bad, to understand why it is bad to use a collection in which we clearly know how many elements, or a collection in which we can access an element by index. I just go for general reasons that it is better and prevents unnecessary calculations in the future. I also walk out of considerations that right now this is actually `ReadOnlyCollection` anyway. On the other hand, I don’t remember the index call.

dedok1997:
I merge my changes with yours. And now there isn't NP exception

rstm-sf:
Thanks for watching!

ForNeVeR:
@dedok1997, before merging, please do the following:

1. Rebase your branch on top of the current `master`
2. Update the serialization code to ignore the `Atom.Children` property. I don't think it should ever be serialized in tests.
   
   I believe that the best way to fix that is to filter properties via our custom `InnerPropertyContractResolver` (probably it would need to be renamed after that).

src/WpfMath/Atoms/Atom.cs
ForNeVeR: I would prefer this property to be `virtual` to not pass the implementation via the `Atom`'s constructor.

src/WpfMath/Atoms/RowAtom.cs
ForNeVeR: I don't like this cast. Let's make it the other way around? Leave poor `Elements` as-is and implement the new `Children` property via `Elements`:
```
public override IEnumerable<Atom> Children => Elements;
```

src/WpfMath/TexFormulaHelper.cs
```diff
@@ -153,9 +153,34 @@ public void AddOperator(string operatorFormula, string lowerLimitFormula, string
                 formulaParser.Parse(upperLimitFormula), useVerticalLimits);
         }
 
+        public int MoveOffset(Atom atom, int pos)
```
ForNeVeR: This method hasn't the best name (probably `RecursivelyChangeSourceOffset` would sound better?). Also, should it be `private`?

We definitely need a comment with a motivation (why do we ever need to change a source offset? Because of the predefined formulae, of course).

Also, I think it could be marked as `internal` and tested separately? Would it make sense?

src/WpfMath/TexFormulaHelper.cs
```diff
@@ -153,9 +153,34 @@ public void AddOperator(string operatorFormula, string lowerLimitFormula, string
...
+        {
+            if (atom.Source != null)
+            {
+                atom.Source = new SourceSpan(atom.Source.Source, pos, atom.Source.Length);
```
ForNeVeR: Please leave a thorough comment about what're you doing here. Every piece of code that deals with sources in this way (rebasing sources to other place in formula) demands a detailed comment on what's going on.

src/WpfMath/TexFormulaHelper.cs
```diff
@@ -153,9 +153,34 @@ public void AddOperator(string operatorFormula, string lowerLimitFormula, string
...
+            return pos;
+        }
+
         public void AddOperator(string operatorFormula, bool useVerticalLimits)
         {
-            AddOperator(formulaParser.Parse(operatorFormula), null, null, useVerticalLimits);
+            var formula = formulaParser.Parse(operatorFormula);
+            if (formula?.RootAtom.Source != null)
+            {
+                formula.RootAtom.Source = new SourceSpan(source.Source, source.Start, formula.RootAtom.Source.Length);
```
ForNeVeR: Why wouldn't you use `MoveOffset(formula.RootAtom, source.Start + 1)` here? Would that work?

src/WpfMath/TexFormulaHelper.cs
```diff
@@ -153,9 +153,34 @@ public void AddOperator(string operatorFormula, string lowerLimitFormula, string
...
+                var pos = source.Start + 1;
```
ForNeVeR: Honestly I don't getting why this `+ 1` is here. Please leave a comment about that.

src/WpfMath/Atoms/Atom.cs
ForNeVeR: I also thought that more atoms deserve special `Children` processing: every atom that has some children kept in its properties should override/provide the `Children` property. Otherwise, it won't work well.

Currently, you've only implemented it in the `RowAtom`, this isn't enough.

src/WpfMath/TexFormulaParser.cs
```diff
@@ -609,7 +609,15 @@ private Color ReadColorModelData(SourceSpan value, ref int position)
                 // Predefined formula was found.
                 var predefinedFormula = factory(formulaSource);
                 var atom = AttachScripts(formula, value, ref position, predefinedFormula.RootAtom, true, environment);
-                formula.Add(atom, formulaSource);
+                if (atom.Source != null)
+                {
+                    var source = new SourceSpan(formulaSource.Source, formulaSource.Start, predefinedFormula.RootAtom.Source.Length + 1);
```
ForNeVeR: Would `commandSpan` work here? If not, then please leave a comment about what're you trying to achieve here.

ForNeVeR:
Okay, I'm taking over this PR and will perform the remaining work.

dedok1997:
If it needs to do it right now, ok. But I can do it, later and explain my solution in comments

ForNeVeR:
It just turned out to be on my schedule right now, when I have both the time and the willing to finish this particular task. So, don't worry, I'll make it :)

Thanks, you've done good work.

dedok1997:
Ok, I leave some comments about my code

ForNeVeR:
There's no need for it. I've read all the code and I believe I currently understand what it does. I have some thoughts on tweaking the result a bit, but the core solution will stay the same.

ForNeVeR:
@dedok1997, thanks a lot for your contribution, I appreciate it much. After thorough consideration and discussions summarized in #151, I've decided to look at this problem from another angle, see #246.

That PR supercedes this one, so I'll close it.

PR #219, opened 2019-10-22
Upgrade to .NET Framework 4.6.1, add .NET Core 3.0 target

hez2010:
1. Some tests failed due to float-point precision (Microsoft has upgrade their float-point number IEEE standard), please fixes them after merging this pull request.
2. .NET Framework 4.6.1 as minimal version is required to target both .NET Framework and .NET Core.
3. Only support Visual Studio 2019, because Visual Studio 2017 has no build target file for Sdk `Microsoft.NET.Sdk.WindowsDesktop`.

- Closes #208
- Closes #209 
- Closes #217

## TODO (by @ForNeVeR)

- [ ] Make sure the tests are green
- [ ] Review the NuGet packaging instructions
- [ ] Verify that #217 is really fixed in VS
- [ ] Final review
- [ ] Create new issues for the remaining `TODO`s

PR #228, opened 2019-12-08
WPF support for .NET Core 3.0

### LINKED ISSUE #208: New languages migration
Guess what time it is? Yes, it's _that_ time.

We should migrate to C# 8 and F# 4.7 (while not losing compatibility with all the older frameworks we're currently compatible with).

C# 8 migration is currently blocked on #205, and F# 4.7 migration is blocked on the support in the tooling I use/develop. Both issues will be resolved soon.

### LINKED ISSUE #209: WPF support on .NET Core 3.0
We have to support the new WPF that's released for .NET Core 3.0.

For now, I'm not sure what we should do to support it, but we definitely want to do it.

### LINKED ISSUE #217: Test failure in VS2019
After being able to overcome #210, I've encountered a test failure:
image

It looks like VS' test runner uses some slightly outdated (or too modern?) .NET version than we require in our tests.

I believe we should somehow fix that by adding a specialized double serializer for tests. The numbers in question, 0.52778099999999994 and 0.527781, are **exactly equal** (at least my current pwsh session tells me that). We could try to experiment with `.ToString("r")` to use a round-trippable format or something like that. Looks like we unnecessarily expect more precision than is required by the modern approach to double serialization.

### LINKED ISSUE #226: Add TFM into nuspec
See [here](https://docs.microsoft.com/en-us/nuget/reference/errors-and-warnings/nu5128): we could probably either use `dotnet pack` for nuget package generation or update our manually-written nuspec to include that.

Also, it would be nice to generate the package on the CI server, and to automatically upload it to GitHub.

ForNeVeR:
This PR is based on #219, but I had to create a new PR because I want to force-push some infrastructure upgrades and I have no permission to force-push to the `master` branch of @hez2010's repository.

So, we'll migrate from Sunburst SDK to a .NET Core native one, but will still support .NET 4.5.2+ with .NET Core 3.0.

I had to remove `System.ValueTuple` package because of the troubles it creates for multitarget projects.

Also, I had to fix #217 by introducing more robust `double` formatter because of differences between frameworks we can't not tolerate (because both the frameworks use the same test data).

Appveyor will use `Visual Studio 2019 Preview` image for now, because current `Visual Studio 2019` has no .NET Core 3.1 support yet. We'll likely replace it with `Visual Studio 2019` later, when they'll update their image.

- Closes #208
- Closes #209
- Closes #217
- Closes #226

## TODO

- [x] Make sure the tests are green
- [x] Review the NuGet packaging instructions
- [x] Check if we could fix #226 here, too
- [x] Verify that #217 is really fixed in VS
- [x] Final review
- [x] Create new issues for the remaining `TODO`s
- [x] Create an issue to enable `#nullable` everywhere in the C# project


ForNeVeR:
@hez2010, thanks a lot for your contribution!

PR #231, opened 2019-12-16
Remove some duplicated code from the JSON converters

ForNeVeR:


ForNeVeR:
Thanks for your review!

PR #233, opened 2019-12-23
add ComboBox into an example application

### LINKED ISSUE #179: Add some sort of combobox into an example application
Looking at the code [like that](https://github.com/ForNeVeR/wpf-math/pull/155/files#diff-ab98873d0f2f4d4adf4a8b82dd30a706R101) in our example app, I start to think that we want some kind of combobox to choose among these example formulae.

NikolaiSviridov:
Closes #179.

### DIFF
--- a/src/WpfMath.Example/MainWindow.xaml.cs
+++ b/src/WpfMath.Example/MainWindow.xaml.cs
@@ -33,8 +33,8 @@ namespace WpfMath.Example
         {
             InitializeComponent();
 
-            comboBox.ItemsSource = _testFormulas;
-            comboBox.SelectedIndex = 0;
+            FormulaSelector.ItemsSource = _testFormulas;
+            FormulaSelector.SelectedIndex = 0;
         }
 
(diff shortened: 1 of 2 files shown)


src/WpfMath.Example/MainWindow.xaml.cs
```diff
@@ -119,5 +123,11 @@ private void inputTextBox_SelectionChanged(object sender, RoutedEventArgs e)
             formula.SelectionStart = inputTextBox.SelectionStart;
             formula.SelectionLength = inputTextBox.SelectionLength;
         }
+
+        private void ComboBox_OnSelectionChanged(object sender, SelectionChangedEventArgs e)
+        {
+            string text = (sender as ComboBox).SelectedItem as string;
```
ForNeVeR:
```suggestion
            string text = comboBox.SelectedItem as string;
```

ForNeVeR:
@NikolaiSviridov, thanks for your help!

PR #236, opened 2020-01-02
Destroy the BOM

### LINKED ISSUE #131: Fix BOM in files
It looks like VS2017 finally started to remove BOM according to our `.editorconfig` file. That causes some noise in the incoming commits.

I think that we need to confirm that and, if that's actually true, remove the BOM's everywhere.

ForNeVeR:
This PR removes UTF-8 BOM from all the files in the repository (where it was accidentally added), and also adds a validation to CI so nobody will be able to commit more BOMs.

The aBOMination will be stopped!

(I've verified that the latest VS2019/.NET Core SDK/Rider all work well with the sources encoded in UTF-8 without BOM.)

Closes #131.

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -1,9 +1,10 @@
 image:
     - Visual Studio 2019
+before_build:
+    - pwsh: ./scripts/verify-encoding.ps1
 build_script:
     - dotnet build
 after_build:
-    - pwsh: ./scripts/verify-encoding.ps1
     - dotnet pack
 test_script:
     - dotnet test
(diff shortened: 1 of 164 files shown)


appveyor.yml
```diff
@@ -1,8 +1,9 @@
-﻿image:
+image:
     - Visual Studio 2019
 build_script:
     - dotnet build
 after_build:
+    - pwsh: ./scripts/verify-encoding.ps1
```
ForNeVeR: Let's put in (and any such validation in future) into `before_build`.

PR #240, opened 2020-01-10
Fix throw exception for the non-ascii symbol (#225)

### LINKED ISSUE #225: \text doesn't work with indices if there's only one Cyrillic letter
Consider this sample:
```
\text{А}_{\text{И}}
```

image

At the same time, the following works:

```
\text{АБ}_{\text{И}}
```

image

rstm-sf:
Fixes #225

### DIFF
--- a/src/WpfMath/Atoms/ScriptsAtom.cs
+++ b/src/WpfMath/Atoms/ScriptsAtom.cs
@@ -81,7 +81,8 @@ namespace WpfMath.Atoms
                 shiftUp = resultBox.Height - texFont.GetSupDrop(superscriptStyle.Style);
                 shiftDown = resultBox.Depth + texFont.GetSubDrop(subscriptStyle.Style);
             }
-            else if (this.BaseAtom is CharSymbol charSymbol)
+            else if (this.BaseAtom is CharSymbol charSymbol
+                     && charSymbol.IsSupportedByFont(texFont, style))
             {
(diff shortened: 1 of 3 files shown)


src/WpfMath/Atoms/ScriptsAtom.cs
```diff
@@ -81,7 +81,8 @@ protected override Box CreateBoxCore(TexEnvironment environment)
                 shiftUp = resultBox.Height - texFont.GetSupDrop(superscriptStyle.Style);
                 shiftDown = resultBox.Depth + texFont.GetSubDrop(subscriptStyle.Style);
             }
-            else if (this.BaseAtom is CharSymbol charSymbol)
+            else if (this.BaseAtom is CharSymbol charSymbol
+                     && charSymbol.IsSupportedByFont(texFont, style))
```
rstm-sf: Maybe it should be inside the scope? I could not figure out what the ```shiftUp``` and ```shiftDown``` variables should be equal in this case

ForNeVeR: For now, I believe this is okay, because we literally don't know the `supdrop` parameter of the ordinary font. For TeX fonts, this is a parameter that's defined by font metrics.

I'll create a new umbrella issue about guessing these parameters for system fonts, and add this as a motivating example there.

PR #145, opened 2018-08-15
Implemented the underline command

B3zaleel:

PR #242, opened 2020-01-11
Merge 'master' into '17-avalonia'

rstm-sf:
Hello!

Merge 'master' into '17-avalonia'

Conflicts resolved with _kdiff3 Version 0.9.98 (64 bit)_

Here I did not understand what is better to choose and left as is:

https://github.com/ForNeVeR/wpf-math/blob/36707619581933b6c9fef213627fdb7beb24236b/src/WpfMath/TexFontInfo.cs#L143

https://github.com/ForNeVeR/wpf-math/blob/fd5c3b4433bfafe434ec282272adf00c4a3b6f7c/src/WpfMath/TexFontInfo.cs#L143


rstm-sf:
Sorry, but the rest I do not know how to update. For example, how to add this?

https://github.com/ForNeVeR/wpf-math/blob/36707619581933b6c9fef213627fdb7beb24236b/src/WpfMath/Rendering/IElementRenderer.cs#L42-L50

rstm-sf:
Now I thought that maybe it would be possible to transfer to the _shared_ `Parsers` and do something with `Colors`

ForNeVeR:
> Here I did not understand what is better to choose and left as is:
> 
> https://github.com/ForNeVeR/wpf-math/blob/36707619581933b6c9fef213627fdb7beb24236b/src/WpfMath/TexFontInfo.cs#L143
> 
> https://github.com/ForNeVeR/wpf-math/blob/fd5c3b4433bfafe434ec282272adf00c4a3b6f7c/src/WpfMath/TexFontInfo.cs#L143

In general, you should prefer the code version from `master` when merging `master` to other branches.

Here, we have an actual error in `master` (this method never returns `null` in the modern code base), so I'll correct it separately :)

ForNeVeR:
> Sorry, but the rest I do not know how to update. For example, how to add this?
> 
> https://github.com/ForNeVeR/wpf-math/blob/36707619581933b6c9fef213627fdb7beb24236b/src/WpfMath/Rendering/IElementRenderer.cs#L42-L50

This stuff is used to properly process background colors in `\colorbox`: namely, this is the issue #128.

For now, you could leave this method empty in Avalonia renderer counterpart (but please add it to the `IElementRenderer` interface if it is duplicated for Avalonia). I'll add that issue to the work list.

ForNeVeR:
> Now I thought that maybe it would be possible to transfer to the _shared_ `Parsers` and do something with `Colors`?

Yes, it should be possible to make all the parser stuff shared. And, yeah, color support is a bit problematic, because WPF and Avalonia have different color types. Any ideas?

rstm-sf:
I do not understand what is wrong. I have the solution building success

rstm-sf:
What? Ok

src/AvaloniaMath/TexFormulaParser.cs
```diff
@@ -64,27 +77,23 @@ internal static string[][] DelimiterNames
 
         private static void Initialize()
         {
-            
-            commands = new HashSet<string>
+            /* TODO: What?
+            //
+            // If start application isn't WPF, pack isn't registered by defaultTexFontParser
```
rstm-sf: Could you tell me what this means?

src/AvaloniaMath/TexFormulaParser.cs
```diff
@@ -363,80 +529,108 @@ private static SourceSpan ReadElement(SourceSpan value, ref int position)
...
+                            environment.CreateChildEnvironment());
 
                         source = value.Segment(start, position - start);
-                        return new Radical(source, sqrtFormula.RootAtom, degreeFormula?.RootAtom);
+                        return new Tuple<AtomAppendMode, Atom>(
+                            AtomAppendMode.Add,
+                            new Radical(source, sqrtFormula.RootAtom, degreeFormula?.RootAtom));
                     }
+                /* TODO: What?
                 case "underline":
```
rstm-sf: It appeared in #145, but now not in master

rstm-sf:
@ForNeVeR, I ended up upgrading to the current master

rstm-sf:
It seems like I did something wrong

pic
pic

rstm-sf:
> It seems like I did something wrong (this happened in today's commits...)

This is due to the command
```tex
\left{ \right}
```

rstm-sf:
Why? It's work

picture

src/AvaloniaMath/TexFormulaParser.cs
rstm-sf: It's here https://github.com/ForNeVeR/wpf-math/blob/0ed0a3b1c32a0ac5784ec6747cbd454242b6d592/src/WpfMath/Parsers/StandardCommands.cs#L105

src/AvaloniaMath/Colors/PredefinedColorParser.cs
```diff
@@ -0,0 +1,76 @@
...
+                if (alphaFraction == null || alphaFraction < 0.0 || alphaFraction > 1.0)
+                    return null;
+
+                alpha = (byte) Math.Round(255.0 * a, MidpointRounding.AwayFromZero);
+            }
+
+            if (alpha == null)
+                return null;
+
+            color = Color.FromArgb(alpha.Value, color.R, color.G, color.B);
```
rstm-sf: > Copy-paste to Avalonia/Colors from WPF/Colors

Sorry, but I have no idea but to do this. It differs only in this line (without `Avalonia.Media`)

rstm-sf: You can still try the simple way using preprocessor directives and the fact that the project for avalonia is built under netstandard

rstm-sf: > because no setters

[readonly struct](https://github.com/AvaloniaUI/Avalonia/commit/d55edaa18a2791ff024491f25820022c1d1a8531#diff-347257ebc1fd2f1d8dd665dec376da17)

src/AvaloniaMath/TexFormulaParser.cs
ForNeVeR: In WPF application startup code, the `pack://` URL is registered. I believe this is required for us to properly render the fonts in a non-WPF environment (in cases when people use WPF-Math in ASP.NET apps, WinForms or something like that: e.g. in our unit tests, I think).

I don't think it's necessary in Avalonia code. How did it ended up there, anyway? Was this piece of code changed lately?

src/AvaloniaMath/Colors/PredefinedColorParser.cs
ForNeVeR: It won't work via preprocessor, because of how .NET compilation model is organized. If we want to leave _one_ version of the common library, then we'll need to compile it _once_, with exactly _one_ set of the preprocessor definitions.

I'll think about it, probably a bit of generic magic/parametrization will help us here.

src/AvaloniaMath/TexFormulaParser.cs
rstm-sf: Thanks! No, there were no changes. I will then delete it

src/AvaloniaMath/Colors/PredefinedColorParser.cs
rstm-sf: > If we want to leave _one_ version of the common library, then we'll need to compile it _once_

Sorry, I have no experience, so I did not think about this

ForNeVeR: You know what? Let's leave it as-isto not complicate the merge. Generalization will be performed on another stage of work.

ForNeVeR:
I have cleaned up the branch a bit by annihilating an unnecessary commit/revert pair.

ForNeVeR:
@rstm-sf, thanks for help!

PR #243, opened 2020-01-12
Remove a misleading comment

ForNeVeR:
The issue with this comment was noticed by @rstm-sf in #242.

This method never returns `null` (or `Result.Ok(null)`) and will return a `Result.Error` instead. I don't think that the comment without a remark about `null` holds any actual value, so I've decided to remove it completely.


ForNeVeR:
Thanks for review!

PR #246, opened 2020-01-19
Source name and source checks in selection renderer

### LINKED ISSUE #132: Fix atom Source assignment in TexFormulaHelper
See this TODO: 

https://github.com/ForNeVeR/wpf-math/blob/1d1297b25ac72f19365221b71b739ddcf3a42182/src/WpfMath/TexFormulaHelper.cs#L17-L19

### LINKED ISSUE #151: SourceSpan interference

ForNeVeR:
Closes #132.
Closes #151.

### DIFF
--- a/src/WpfMath/SourceSpan.cs
+++ b/src/WpfMath/SourceSpan.cs
@@ -4,8 +4,9 @@ namespace WpfMath
 {
     public class SourceSpan : IEquatable<SourceSpan>
     {
-        public SourceSpan(string source, int start, int length)
+        public SourceSpan(string sourceName, string source, int start, int length)
         {
+            SourceName = sourceName;
             this.Source = source;
             this.Start = start;
             this.Length = length;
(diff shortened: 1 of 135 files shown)


src/WpfMath/SourceSpan.cs
```diff
@@ -21,16 +22,22 @@ public SourceSpan(string source, int start, int length)
 
         public string Source { get; }
 
+        /// <summary>A name identifying the source (e.g. a user input, or some sort of macro body).</summary>
+        public string SourceName { get; }
```
gsomix: It doesn't clear what names are allowed here. Is it finite set of names and could be described as enum instead, or something more generic? Please clarify.

ForNeVeR: My intention was to make any names allowed: user may pass them directly to the parser. They have purely informative function.

ForNeVeR:
Thanks for review!

PR #249, opened 2020-01-25
Fix parameter order in ArgumentException constructor

ForNeVeR:
Thanks to [Iλya Che in DotNetRuChat](https://t.me/DotNetRuChat/752883) for noticing the issue.


ForNeVeR:
Thanks for review!

PR #252, opened 2020-02-08
added support for \{ and \}

sskodje:
Added support for \\{ and \\} in addition to the existing \lbrace and \rbrace. These commands are in standard latex, but was missing here, making it incompatible.

### DIFF
--- a/src/WpfMath/Data/DefaultTexFont.xml
+++ b/src/WpfMath/Data/DefaultTexFont.xml
@@ -110,7 +110,7 @@ for a text style is not defined (e.g. "numbers" and "small" in "mathcal") -->
     <SymbolMapping name="widehat" ch="98" fontId="2"/>
     <SymbolMapping name="widetilde" ch="101" fontId="2"/>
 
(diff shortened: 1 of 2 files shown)


src/WpfMath/Data/DefaultTexFont.xml
```diff
@@ -110,8 +110,9 @@ for a text style is not defined (e.g. "numbers" and "small" in "mathcal") -->
     <SymbolMapping name="widehat" ch="98" fontId="2"/>
     <SymbolMapping name="widetilde" ch="101" fontId="2"/>
 
-    <!-- delimiters that can change size -->
-
+    <!-- delimiters that can change size -->Q
```
ForNeVeR: That `Q` shouldn't be here, right? :)

ForNeVeR:
Could you please fix the minor issue I've found? I'll merge this afterwards.

Thanks for your help!

sskodje:
Whoops, that was my fat fingers. Fixed it now.

PR #253, opened 2020-02-09
Added equal padding to all sides when saving to bitmap

sskodje:
The RenderToBitmap(double x, double y) method only adds padding to the left and top side of the bitmap. This fix adds padding to bottom and right as well. Perhaps there should be a Thickness as input instead though? It would make more sense but is a breaking change.


ForNeVeR:
Thank you! Yes, this change makes sense.

I would be a bit sceptical about changing that to `Thickness` since this is a WPF-specific type, and we're trying to be as WPF-agnostic as possible now, because we want to support both WPF and Avalonia in future.

At least let's try to not add WPF dependencies to our public API without very good reasons.

PR #254, opened 2020-02-10
Fix for scripts with curly braces after a command with  curly braces,…

sskodje:
Commands like e.g \hat{x}_{y} currenly throws an exception, but it should be valid latex. I've tested this change for a little while and have not encountered any problems.


ForNeVeR:
Thanks!

PR #255, opened 2020-02-15
Fixed setting foreground color for formula control

sskodje:
Currently, setting the Foreground property of FormulaControl does not work as expected. This fix renders the formulas in the color of the Foreground property.


ForNeVeR:
Thanks for your contribution! This PR has been superseded by #287; the underlying issue is fixed.

PR #256, opened 2020-02-16
Refactor color parsers for separated wpf and avalonia

rstm-sf:
Hello!

Refactor color parsers for separated wpf and avalonia.

Note: dotnet/wpf#2274

**TODO[F]:**

- [x] Check out how `RgbaColor` ctor is called
- [x] Get rid of "_Extract create value for Color to helper for:_" commits
  - _whatever, let them live_
- [ ] Perform a final review iteration


rstm-sf:
It might be nice to make a breaking change and change the `IColorParser` interface:
```c#
Color? Parse(IEnumerable<string> components);
```
to
```c#
Color? Parse(IReadOnlyList<string> components);
```

ForNeVeR:
Go ahead and do it, I think that's an acceptable change.

rstm-sf:
Somehow I don't like the last change

rstm-sf:
Can we use partial for the rest of the code?

rstm-sf:
And what todo for this?
> Note: dotnet/wpf#2274

rstm-sf:
Hmm, or is it better to create a Color structure and use it until we need to draw?

ForNeVeR:
@rstm-sf 
> Can we use partial for the rest of the code?

What exactly would you like to do `partial` here?

> And what todo for this?
> > Note: dotnet/wpf#2274

As I understand, current workaround is to either downgrade to .NET Core SDK 3.1 (from 3.1.1 you have from VS' update, I presume?) or wait for 3.1.2 to release. According to [this comment](https://github.com/dotnet/wpf/issues/2274#issuecomment-580387496), it should be released tomorrow.

> Hmm, or is it better to create a Color structure and use it until we need to draw?

I believe that struct would be better than tuple since we cannot use value tuples in our code due to [a bug in the SDK](https://github.com/dotnet/sdk/issues/3770) :(

rstm-sf:
> I believe that struct would be better than tuple since we cannot use value tuples in our code due to [a bug in the SDK](https://github.com/dotnet/sdk/issues/3770) :(

Perhaps this is somehow connected with failed build

rstm-sf:
> What exactly would you like to do `partial` here?

All that is not specific to ui-framework-project, but the rest is shared. Is it possible?

rstm-sf:
Although, you can try `RgbaColor` to declare `public` and do without `partial`

rstm-sf:
Some of such a change. Then we will not need helper in this form

rstm-sf:
Or maybe it's better to do it with SkiaSharp? Or is it better in the form of a plan for the future?

rstm-sf:
I thought of writing something like this to sharing `TexFormulaParser`:

```c#
public StyledAtom(SourceSpan source, Atom atom, RgbaColor? backgroundColor, RgbaColor? foregroundColor)
  : this(source, atom, GetSolidColorBrush(backgroundColor), GetSolidColorBrush(foregroundColor))
{ }
```

But met it:
https://github.com/ForNeVeR/wpf-math/blob/c051e6b74eddfdece57989fcfc4aad39b60d8d9e/src/WpfMath/TexFormulaParser.cs#L82

rstm-sf:
If not mistaken, then `RgbaColor` cannot be declared as `internal` because of these lines

https://github.com/ForNeVeR/wpf-math/blob/c051e6b74eddfdece57989fcfc4aad39b60d8d9e/src/WpfMath/TexFormulaParser.cs#L151-L154

rstm-sf:
If you want, then I can start changing this branch again. I just don’t know how to get rid of extra commits in another way.

src/SharedMath/Colors/RgbaColor.cs
```diff
@@ -0,0 +1,25 @@
...
+
+        public byte B { get; set; }
+
+        public byte A { get; set; }
+
+        public RgbaColor(byte r, byte g, byte b)
+            : this(0xFF, r, g, b)
+        {}
+
+        public RgbaColor(byte a, byte r, byte g, byte b)
```
ForNeVeR: Not sure it is good idea to call class `RgbaColor` and then accept components in ARGB order :)

Could you please update that to use RGBA order? Then you'll be able to get rid of second constructor using the default argument value.

ForNeVeR:
@rstm-sf 

> > What exactly would you like to do `partial` here?
> 
> All that is not specific to ui-framework-project, but the rest is shared. Is it possible?

I am still not sure what exactly you're proposing. Considering we want to eventually get rid of `.shproj` hell, I don't think it would be possible.

> Although, we can try `RgbaColor` to declare `public` and do without `partial` (or or better to call WpfMathColor :))

I quite like the name already. I don't think we need more "WPF" in our portable class names :)

And, yeah, it should be public, because we allow our users to add their own color parsers.

> Or maybe it's better to do it with SkiaSharp? Or is it better in the form of a plan for the future?

Are you proposing to use SkiaSharp base color instead of our own `RgbaColor`? I don't think it's a good idea: we don't need the SkiaSharp in our WPF parts, and we shouldn't need it.

> I thought of writing something like this to sharing `TexFormulaParser`:
> 
> ```c#
> public StyledAtom(SourceSpan source, Atom atom, RgbaColor? backgroundColor, RgbaColor? foregroundColor)
>   : this(source, atom, GetSolidColorBrush(backgroundColor), GetSolidColorBrush(foregroundColor))
> { }
> ```
> 
> But met it:
> https://github.com/ForNeVeR/wpf-math/blob/c051e6b74eddfdece57989fcfc4aad39b60d8d9e/src/WpfMath/TexFormulaParser.cs#L82

This is a good idea, let's do it in a separate PR. Also, I can't see how WPF resource stuff interferes with this part of your plan.

For the static initializer itself, this is a workaround about `pack:` scheme registration in non-WPF applications (e.g. web apps or unit tests); there're our users who need that.

We could fix that issue by adding a separate scheme registration step and keeping it in the WPF library only. I don't think we'll forget to do that, considering we use this code for our own unit tests. So, this problem will be solved during the consolidation step of Avalonia/WPF sources (when we'll be merging `AvaloniaMath/TexFormulaParser.cs` and `WpfMath/TexFormulaParser.cs` together). You shouldn't care too much about that piece of code right now.

> I just don’t know how to get rid of extra commits in another way.

Which commits do you consider as "extra"? I'm not sure what are you proposing in this comment.

src/SharedMath/Colors/RgbaColor.cs
rstm-sf: For some reason, at one point it seemed to me a good idea to make order as in wpf/avalonia

rstm-sf:
> Considering we want to eventually get rid of `.shproj` hell

Oops, I thought that the idea of ​​using _shproj_ would continue to be preserved.



> Which commits do you consider as "extra"?

Commits by _"Extract create value for Color to helper for:"_ turned out to be unnecessary

src/SharedMath/Colors/RgbaColor.cs
ForNeVeR: Ah, whatever, I like RGBA better :)

rstm-sf:
> For the static initializer itself, this is a workaround about `pack:` scheme registration in non-WPF applications (e.g. web apps or unit tests); there're our users who need that.

Sorry, but I don’t understand what you mean by that. I have no knowledge about this

ForNeVeR:
> I thought that the idea of ​​using _shproj_ would continue to be preserved.

I'm sorry I haven't informed you of these plans.

`.shproj` would be very problematic for combined applications. Consider that somebody wants to use _both_ AvaloniaMath and WpfMath in the same application. They would end up in total mess from clashed class names. This would be **unacceeeptable**!

> Commits by _"Extract create value for Color to helper for:"_ turned out to be unnecessary

Okay, I'll do something with them on merge :)

> > For the static initializer itself, this is a workaround about `pack:` scheme registration in non-WPF applications (e.g. web apps or unit tests); there're our users who need that.
> 
> Sorry, but I don’t understand what you mean by that. I have no knowledge about this

In other words: if you remove this static initializer, WPF-based tests will fail. Is it enough reason to keep this code block for now? :)

rstm-sf:
> In other words: if you remove this static initializer, WPF-based tests will fail. Is it enough reason to keep this code block for now? :)

I realized that this is needed to work somewhere else. I do not understand what can be done so that wpf/avalonia code can be merged

ForNeVeR:
Well, the current task you're solving is to extract the `Color` stuff, right? I think this PR does that job good. Remaining issues for now could be left for next iteration of work.

rstm-sf:
> Well, the current task you're solving is to extract the `Color` stuff, right?

Yes, I just started with functions, but ended with classes and settled on `TexFormulaParser`. I can stop here. I’m just saying that I don’t understand how to move on in the merging :)

ForNeVeR:
> I’m just saying that I don’t understand how to move on in the merging :)

I have no definite answer for that, but we'll need to improve our API to perform this initialization somewhere. We'll decide later. It'll be fine.

PR #259, opened 2020-02-29
Replace bg/fg Brush with RgbaColor

rstm-sf:
Replace bg/fg Brush with RgbaColor

It seems that we only need to know the color information:
- [Using colours in LaTeX](https://www.overleaf.com/learn/latex/Using_colours_in_LaTeX)
- [Brushes and text](https://docs.microsoft.com/en-us/windows/uwp/design/style/brushes#brushes-and-text)


src/AvaloniaMath/Rendering/AvaloniaElementRenderer.cs
```diff
@@ -49,16 +52,17 @@ public void RenderGlyphRun(Func<double, GlyphRun> scaledGlyphFactory, double x,
...
+
+                _drawingContext.DrawText(brush, origin, ft);
             }
         }
 
-        public void RenderRectangle(Rect rectangle, IBrush foreground)
+        public void RenderRectangle(Rect rectangle, RgbaColor? foregroundColor)
         {
+            // TODO: foregroundColor can be null?
+            var brush = new SolidColorBrush(foregroundColor.Value.ToAvaloniaColor());
```
rstm-sf: Does it need nullable or non-nullable here?

src/WpfMath/Controls/FormulaControl.xaml.cs
```diff
@@ -125,7 +127,7 @@ private void Render()
             var visual = new DrawingVisual();
             var renderer = texFormula.GetRenderer(TexStyle.Display, Scale, SystemTextFontName);
 
-            var selectionBrush = SelectionBrush;
+            var selectionBrush = (SolidColorBrush)SelectionBrush;
```
rstm-sf: This is right? By the way, I did not find this for Avalonia

src/SharedMath/Colors/RgbaColor.cs
```diff
@@ -17,5 +22,14 @@ public RgbaColor(byte r, byte g, byte b, byte a = 0xFF)
             this.B = b;
             this.A = a;
         }
+
+        public uint ToUint32() =>
+            ((uint)A << 24) | ((uint)R << 16) | ((uint)G << 8) | (uint)B;
```
rstm-sf: And I began to understand why they call ARGB. Can we change the sequence and add a parser from the hex format for the converter?

src/AvaloniaMath/Rendering/AvaloniaElementRenderer.cs
rstm-sf: Maybe: ```foregroundColor.HasValue ? foregroundColor.Value.ToAvaloniaColor() : null```

src/WpfMath/Controls/FormulaControl.xaml.cs
rstm-sf: Probably better
```c#
var objColor = SelectionBrush
            .GetType()
            .GetField(nameof(Color))
            .GetValue(null);
if (objColor is Color selectionColor)
```

rstm-sf: But then, it’s better to set `Color` instead of a `Brush` immediately in the control properties

rstm-sf: > This is right? By the way, I did not find this for Avalonia

fix to pattern matching, because it was a bug

src/AvaloniaMath/Rendering/AvaloniaElementRenderer.cs
ForNeVeR: I would make it non-nullable, until we find a case when `null` is useful.

src/WpfMath/Controls/FormulaControl.xaml.cs
ForNeVeR: It is better to replace `Brush` with our `ArgbColor` abstraction if we don't support any non-solid color brushes anymore.

src/SharedMath/Colors/RgbaColor.cs
ForNeVeR: I don't see why you've decided to use ARGB instead of RGBA, but whatever, let's change it if you think it's better this way. But why exactly?

ForNeVeR:
I've skimmed through this draft PR and I think I approve these changes in principle.

This is a breaking change (both technically and conceptually): earlier, it was possible to use any sort of brush to define the color, and now, only solid colors will work.

I don't know if there're any users who require non-solid colors for now, but we'll have to think better about that I think.

WPF is known for its abilities to apply various interesting color and shape transformations to its controls, and it's not bad to lose this ability right now.

I think we could first establish a common code base with a PR like this, and postpone the fully-colored cross-platform solution to a later time point. I'll add the corresponding item to the task list in #17.

src/SharedMath/Colors/RgbaColor.cs
rstm-sf: I can rename it back, for me it is not critical, on the contrary, I even like RGBA more. Just wpf / avalonia represents color as a string in ARGB order (for example, all approved files would have to be edited). Accordingly, we must be able to parse such an order that it will already slightly contradict the previous name of the class. Only on this seemed to me a good idea to do so. I no longer see the benefits of a _custom_ name

src/WpfMath/Controls/FormulaControl.xaml.cs
rstm-sf: The difference will be that the user himself will have to create _knowing_ colors. When, as in wpf / avalonia, there is already a set of colors

rstm-sf:
> WPF is known for its abilities to apply various interesting color and shape transformations to its controls, and it's not bad to lose this ability right now.

Could you clarify what additional features can be used? We can use gradients? Sorry, I do not have enough knowledge and experience in applying colors in either wpf or latex to look at it from the other side. Therefore, I can not find additional properties that we can come in handy

src/AvaloniaMath/Rendering/AvaloniaElementRenderer.cs
rstm-sf: We can forward null to this function and only here make a comparison and get Color.Black from wpf / avalonia

ForNeVeR:
Gradients, textures, [anything that's supported by `Brush`](https://docs.microsoft.com/en-us/dotnet/framework/wpf/graphics-multimedia/wpf-brushes-overview).

ForNeVeR:
Closed in favor of #263.

PR #261, opened 2020-04-03
Fixed crash on empty \sqrt{}

sskodje:
This fixes a crash on an empty square root, like \sqrt{}


ForNeVeR:
Thanks for help! After a brief consideration, I've decided to make another fix: not throwing an exception, but instead rendering such cases (with a radical sign, as https://mathspp.com/texpaste does).

PR #262, opened 2020-04-11
Add Modulo/remainder operator which is in amsmath

GregoryMorse:
This was missing and is a reasonably common mathematics operator, enough to include here for sure.


ForNeVeR (approved):
Thanks!

PR #263, opened 2020-04-20
Replace bg/fg Brush with self-container

rstm-sf:
Hello!

Replace bg/fg Brush with self-container.

This is a different approach than #259, which allows you to keep the original brush


ForNeVeR (approved):
Good to go, thanks for your work!

PR #266, opened 2020-06-30
BigOperator selection

nenaaki:
Hello
The current WpfMath didn't seem to be able to choose \sin \cos \tan etc.
I tried the original of SourceSpan as the Source of BigOperator. Please merge if you like.

### DIFF
--- a/src/WpfMath/TexFormulaHelper.cs
+++ b/src/WpfMath/TexFormulaHelper.cs
@@ -17,7 +17,7 @@ namespace WpfMath
         public TexFormula Formula { get; }
 
         private TexFormula ParseFormula(string source) =>
-            _formulaParser.Parse(new SourceSpan("Predefined formula fragment", source, 0, source.Length));
+            _formulaParser.Parse(new SourceSpan("Predefined formula fragment", source, 0, source.Length), origin: this._source);
 
         public void SetFixedTypes(TexAtomType leftType, TexAtomType rightType)
         {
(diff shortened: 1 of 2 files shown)


ForNeVeR:
Hi!

If I understood you correctly, and we're talking about selection highlighting in formulas: like, if you wrote `\sin{2+2}`, then `\sin` won't highlight correctly.

You seem to resolve this by changing every `CharAtom` inside of a formula instantiation to have its source to be replaced with "origin". I don't think this is the _right_ fix.

See: [here](https://github.com/ForNeVeR/wpf-math/pull/266/files#diff-46b84534570fc41880b00a838b5cc50060ae75e6d7996f90c52b161a0b7aad49R20) we have a "Predefined formula fragment", which serves exactly the purpose of differentiating the formula-internal source and external source.

Say, we have a user formula `\sin{2+2}`. We know (from `PredefinedTexFormulas.xml`) that `\sin` corresponds to `\mathrm{sin}`.

This `\mathrm{sin}` doesn't have any source in the original formula; we have no general way of creating this mapping. Everything inside of a formula was somehow generated by it, but we have no idea how. That's why it don't work.

Your fix works for very simple cases (and even then, it doesn't work completely right: instead of accociating the box around all the characters with the source, it will instead associate every character individually), and will break for more complex cases.

I would lean towards more complex solution which somehow overrides the source for the root formula atom, and not its internal atoms. Also, [this particular line](https://github.com/ForNeVeR/wpf-math/pull/266/files#diff-46b84534570fc41880b00a838b5cc50060ae75e6d7996f90c52b161a0b7aad49R20) is very wrong: on one hand, it tries to differentiate between the original and predefined formula source, and on the other hand, it tries to override the source it has itself defined, by passing `origin: this._source`.

Sorry, but I couldn't accept this contribution, because I don't believe this is the right approach to the problem.

PR #268, opened 2020-07-26
.gitignore cleanup

ForNeVeR:
Two items are getting removed:

- `/packages/` were added in ye olde times, and are only relevant for the old project system.

- `*.DotSettings` is ReSharper/Rider's way of sharing the settings (while any non-shared settings get written into `*.DotSettings.user`), so it shouldn't generally be ignored.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,10 +1,8 @@
 /.idea/
 /.vs/
-/packages/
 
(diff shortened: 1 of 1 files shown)


ForNeVeR:
Thanks for your review!

PR #279, opened 2020-08-23
Fixed warnings about obsolete WPF API in .NET Core.

### LINKED ISSUE #232: Some WPF API is obsolete in .NET Core
Hello!

Some WPF API is obsolete in .NET Core:

- ['GlyphRun.GlyphRun(GlyphTypeface, int, bool, double, IList<ushort>, Point, IList<double>, IList<Point>, IList<char>, string, IList<ushort>, IList<bool>, XmlLanguage)' is obsolete: 'Use the PixelsPerDip override'](https://github.com/ForNeVeR/wpf-math/blob/84fbb40f8024cbbecd1656c64ab89d1a7a3684ad/src/WpfMath/Boxes/CharBox.cs#L30-L32)
- ['FormattedText.FormattedText(string, CultureInfo, FlowDirection, Typeface, double, Brush)' is obsolete: 'Use the PixelsPerDip override'](https://github.com/ForNeVeR/wpf-math/blob/master/src/WpfMath/SystemFont.cs#L118-L123)

```c#
/// <param name="pixelsPerDip">PixelsPerDip of the screen on which this is to be drawn (96ths of an inch).</param>
```

ygra:
Basically the new APIs require one to be explicit about pixels per DIP, which probably helps with high-DPI settings. This patch just gets rid of the warnings in a minimal way, while hopefully retaining the same functionality. It looks identical to me right now in both .NET 4 and .NET Core 3.1. I haven't yet tested with different DPI settings, though.

Intends to fix #232.

### DIFF
--- a/src/WpfMath/Boxes/CharBox.cs
+++ b/src/WpfMath/Boxes/CharBox.cs
@@ -1,3 +1,4 @@
+using System.ComponentModel;
 using System.Linq;
 using System.Windows;
 using System.Windows.Media;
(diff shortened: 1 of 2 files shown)


ForNeVeR (approved):
Thanks for your contribution!

PR #280, opened 2020-08-27
Minor typographical fixes on the example formulas.

ygra:
Some of the example formulas in the sample application are off typographically. From the missing space before dx in an integral to writing words as if they were variables, those are easy fixes, but many people using LaTeX rarely bother.

So, things should be a tad prettier now ;-)


ForNeVeR (approved):
Thanks for help!

PR #284, opened 2020-10-30
Added Atom types to test JSON data.

### LINKED ISSUE #188: Add object types to ApprovalTests.Net data
Currently it could be hard to read the ApprovalTests.Net data because the atom types are ommitted from the JSON. We need to add them.

ygra:
This adds an `[AtomType]` property to all types derived of `Atom`. Don't know any F#, though, so code is probably abysmal.

Fixes #188.

### DIFF
--- a/src/WpfMath.Tests/ApprovalTestUtils.fs
+++ b/src/WpfMath.Tests/ApprovalTestUtils.fs
@@ -16,6 +16,7 @@ open Newtonsoft.Json.Converters
 open Newtonsoft.Json.Serialization
 
 open WpfMath
+open WpfMath.Atoms
 
 type private BomlessFileWriter(data: string, ?extensionWithoutDot: string) =
     inherit ApprovalTextWriter(data, defaultArg extensionWithoutDot "txt")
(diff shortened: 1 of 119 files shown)


src/WpfMath.Tests/ApprovalTestUtils.fs
```diff
@@ -39,11 +39,36 @@ type private InnerPropertyContractResolver() =
...
+        let baseTypes =
+            ``type``
+            |> Seq.unfold (fun t ->
+                    if t.BaseType = null then
+                        None
+                    else
+                        Some(t, t.BaseType)
+                )
+
+        if Seq.exists ((=) typeof<Atoms.Atom>) baseTypes then
```
ForNeVeR, line 58: I guess you wanted just ``` typeof<Atom>.IsAssignableFrom ``type`` ```.

ForNeVeR (changes requested):
Thanks for your contribution! I'll go ahead and simplify the F# part, though it does exactly what I wanted it to :)

ForNeVeR (approved):
Thanks for the contribution! You may wish to look at the resulting code, which really shows where the F# is good at :)

ygra:
Yeah, I was pretty sure my solution wasn't the best one. It's been years that I even looked at F# and I was astonished I even got something to compile (let alone work).

PR #285, opened 2020-11-01
Fixed the link path

hugmouse:
:shrug:


ForNeVeR (changes requested):
As you can see, the real link is declared at the bottom of the document. Could you fix it there?
https://github.com/ForNeVeR/wpf-math/blame/ddeb909d5456c2dd43123087660ec36a260416fe/README.md#L138

Even after your patch, the link at the bottom is defunct.

ForNeVeR (approved):
Thanks!

PR #287, opened 2020-11-08
Support setting foreground and background via WPF control properties

### LINKED ISSUE #99: Foreground property not working on FormulaControl

ForNeVeR:
Closes #99. Obsoletes #162.

I've decided not to keep the binary compatibility with `TexFormula.GetRenderer` signature, since we're still technically in `0.*` version range.


ForNeVeR:
I have reviewed this myself, and believe this is a step in the right direction. It also obsoletes #255.

PR #301, opened 2021-07-13
Move the equals code to shared for `feature/17-avalonia`

rstm-sf:
Move the equals code to shared for `feature/17-avalonia`

I understand that this parser is no longer necessary, for this there are [advanced parsers](https://github.com/ForNeVeR/wpf-math/tree/master/src/WpfMath/Colors) (specifically [`PredefinedColorParser`](https://github.com/ForNeVeR/wpf-math/blob/master/src/WpfMath/Colors/PredefinedColorParser.cs))

https://github.com/ForNeVeR/wpf-math/blob/9e2c2e2a1ad91264f70ad905abbe0bc24cb2f938/src/WpfMath/TexPredefinedFormulaParser.cs#L240-L246


ForNeVeR (approved):
Thanks!

ForNeVeR:
Wait, it's not ready yet?

rstm-sf:
It's done now, I'm confused

> Assembly.GetExecutingAssembly().GetManifestResourceStream(resourceName)

src/AvaloniaMath.Example/App.xaml.cs
```diff
@@ -9,5 +12,18 @@ public override void Initialize()
         {
             AvaloniaXamlLoader.Load(this);
         }
+
+        public override void OnFrameworkInitializationCompleted()
+        {
+            if (ApplicationLifetime is IClassicDesktopStyleApplicationLifetime desktop)
```
ForNeVeR, line 18: Why do you need this type check here? Are you not sure if the lifetime is always of this type?

src/AvaloniaMath.Example/ViewLocator.cs
```diff
@@ -11,17 +11,16 @@ public class ViewLocator : IDataTemplate
...
-                return (Control)Activator.CreateInstance(type);
-            }
-            else
-            {
-                return new TextBlock { Text = "Not Found: " + name };
-            }
+            return type is null
+                ? new TextBlock { Text = "Not Found: " + name }
+                : Activator.CreateInstance(type) is Control control
+                    ? control : new TextBlock { Text = "Not Create Instance: " + type };
```
ForNeVeR, line 23:
```suggestion
                    ? control : new TextBlock { Text = "Created object is not Control: " + type };
```

ForNeVeR (changes requested):

src/AvaloniaMath.Example/App.xaml.cs
rstm-sf: No, I didn't think about that at all. It's just that this code is common in the template

https://github.com/AvaloniaUI/avalonia-dotnet-templates/blob/21fbd4cf2b940801609c7d32f66a14f4f2d97403/templates/csharp/app/App.axaml.cs#L14-L22

ForNeVeR: Alright.

ForNeVeR (approved):
Thanks for help!

PR #303, opened 2021-08-07
C# 9 records

### LINKED ISSUE #265: Consider using C# 9 records for Atom implementations
Earlier, we had too much problems playing with immutable `Atom` type with inheritance, e.g. see #125.

The key problem is that we sometimes want to write something like this:
```csharp
Atom atom = new BigAtom(smth, smth);
Atom atom2 = atom.WithSourceSpan(new SourceSpan(…));
```

This is impossible to write in a world where `Atom` and all its inheritors are immutable. It would be possible to add a virtual `WithSourceSpan` method, but was too much fuss and manual object-cloning code to implement.

In C# 9, as I currently read it, new record classes [will support `with`-expression together with inheritance](https://devblogs.microsoft.com/dotnet/welcome-to-c-9-0/#with-expressions-and-inheritance):

> Value-based equality and non-destructive mutation are notoriously challenging when combined with inheritance. Let’s add a derived record class `Student` to our running example:
> 
> ```csharp
> public data class Person { string FirstName; string LastName; }
> public data class Student : Person { int ID; }
> ```
> And let’s start our with-expression example by actually creating a Student, but storing it in a Person variable:
> 
> ```csharp
> Person person = new Student { FirstName = "Scott", LastName = "Hunter", ID = GetNewId() };
> otherPerson = person with { LastName = "Hanselman" };
> ```
> At the point of that with-expression on the last line the compiler has no idea that `person` actually contains a `Student`. Yet, the new person wouldn’t be a proper copy if it wasn’t _actually_ a `Student` object, complete with the same `ID` as the first one copied over.
> 
> C# makes this work. Records have a hidden virtual method that is entrusted with “cloning” the whole object. Every derived record type overrides this method to call the copy constructor of that type, and the copy constructor of a derived record chains to the copy constructor of the base record. A `with`-expression simply calls the hidden “clone” method and applies the object initializer to the result.

We should keep a close eye on this feature, and try to use it for our purposes after it will be publicly available.

ForNeVeR:
Closes #265.


gsomix (approved):
LGTM.

ForNeVeR:
Thanks for review!

PR #329, opened 2023-01-02
implement CommandParser

rennmaus-coder:
Implemented A Begin CommandParser which parses until the next \end. Doesn't support nested environments.
Should I execlude the other files (fonts and images)?

## TODO[F]
- [x] Add a test.
- [x] Support nested environments.
- [x] More tests:
  - [x] test for `\pmatrix` nested inside of a `\begin{pmatrix}`,
  - [x] tests for every corner case across the code.
- [x] Investigate an example app failure: `\begin{pmatrix}{line 1}{line 2}\end{}`.
- [x] Roll back the changes in `NonRecursiveEnvironment`.
- [x] Documentation:
  - [x] doc-comments,
  - [x] user manual,
  - [x] changelog.

PR #330, opened 2023-01-03
.gitattributes: text=auto

ForNeVeR:
Without this change, files get weirdly mangled on checking them out, `git rebase` and whatnot.

Initially, I got this version of `.gitattributes` from [this comment](https://github.com/actions/checkout/issues/135#issuecomment-613361104) on GitHub `checkout` Action repository.

Why have I even introduced this? Because I've decided we want to keep the line endings as `\n` (otherwise, they themselves get occasionally messed up by myself and other contributors). And then I found that GitHub Actions check out the repository with `autocrlf=true` on Windows. To avoid that, I wanted to do an equivalent of `core.autocrlf=input`, and took the advice from the aforementioned comment.

I was wrong, and I hope the problem gets fixed by this.

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,1 +1,1 @@
-* text eol=lf
+* text=auto eol=lf


ForNeVeR:
An example of how things got messed up you can see in PR https://github.com/ForNeVeR/wpf-math/pull/329: as you can see, `.ttf` and `.png` files become 1 byte lesser, perhaps because a stray `\r\n` was found by Git somewhere there.

Issue #298, opened 2021-07-09
Release: use ChangelogAutomation


ForNeVeR:
Current releases aren't fully automated yet: I have to manually update the `.csproj` file to copy the release notes.

Also, the links in the release notes copied into GitHub release are broken.

Specifically because of these reasons, I've created [ChangelogAutomation](https://github.com/ForNeVeR/ChangelogAutomation) project which integrates well with NuGet release notes, as well as GitHub actions.

ForNeVeR:
Depends on https://github.com/ForNeVeR/ChangelogAutomation/issues/14.

PR #344, opened 2023-01-26
Read changelog uses ForNeVeR/ChangelogAutomation.action@v1

kolosovpetro:
Fixes partially: https://github.com/ForNeVeR/wpf-math/issues/298

What's new:

- Step `Read changelog` uses `ForNeVeR/ChangelogAutomation.action@v1`
- Versions bump for `checkout`, `cache`, `setup-dotnet`
- Added `workflow_dispatch` trigger

PS, I am not able to test nuget push due to security precautions of Github, however, output of read change log file is there: https://github.com/kolosovpetro/wpf-math/actions/runs/4018962182/jobs/6905216105#step:8:11

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -9,28 +9,33 @@ on:
   schedule:
     - cron: '0 0 * * 6'
 
+  workflow_dispatch:
+
 jobs:
   main:
     runs-on: windows-2019
+
     env:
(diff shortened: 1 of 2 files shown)


ForNeVeR (approved):
Thanks!

Notably, the fix is not complete, yet.

To fix this completely, we'd need to also include MSBuild support from ChangelogAutomation.

PR #384, opened 2023-03-11
Change from UserControl to Custom Control.

### LINKED ISSUE #97: Not possible to set default style of FormulaControl in app?
I'm trying to set a default style for `FormulaControl` in my entire app. At the moment it doesn't work, and I believe it might be related to this SO answer:
https://stackoverflow.com/a/39364416/4351604

If i'm correct, this should be added to `FormulaControl.cs`
```
static MyControl()
{
    DefaultStyleKeyProperty.OverrideMetadata(
        typeof(FormulaControl),
        new FrameworkPropertyMetadata(typeof(FormulaControl))); 
}
```

samirem:
This should fix #97.

Instead of inheriting from `UserControl`, the `FormulaControl` now inherits from `Control`, and overrides the `DefaultStyleKey`. I also had to override `MeasureOverride`, and compute the size of the control by making some minor modifications to other classes.


ygra:
You've lost the error template along the way. I think you should be able to include this in a generic.xaml for the assembly, which then provides a default style for the control. The whole concept is a bit underdocumented, admittedly.

src/WpfMath/Controls/FormulaControl.cs
```diff
@@ -0,0 +1,223 @@
...
+        nameof(SystemTextFontName), typeof(string), typeof(FormulaControl),
+        new FrameworkPropertyMetadata(
+            "Arial",
+            FrameworkPropertyMetadataOptions.AffectsMeasure));
+
+    public static readonly DependencyProperty SelectionStartProperty = DependencyProperty.Register(
+        nameof(SelectionStart), typeof(int), typeof(FormulaControl),
+        new FrameworkPropertyMetadata(
+            0,
+            FrameworkPropertyMetadataOptions.AffectsMeasure));
```
ygra, line 101: Does this really change the size of the control? If not, AffectsRender should suffice. Same with SelectionLength and SelectionBrush.

ForNeVeR:
I am so sorry this took me some time to get to review. It somehow slipped through my mailbox. I'll take a closer look in the next few days.

ForNeVeR:
Could you please give an example of how you would like to use/override the default style?

I am concerned by the fact that this introduces several breaking changes into the library, and I'd like to get a code example to play with and maybe figure out if it's possible to do in a less breaking way, or is it worth to introduce a compatibility layer.

ForNeVeR:
Without explanations, I'm sorry, I have to decline this for now. I don't understand the issue, and I don't understand how to check that the issue is resolved.

PR #388, opened 2023-04-24
(#387) Fix alignment issue in matrix with an empty cell.

### LINKED ISSUE #387: Alignment issue in matrix with an empty cell.
There is an alignment problem in the matrix rendering when there is an empty cell.

Ex: `\matrix{A & B \\ A & B \\ & B}`
image

It seems that the issue is located in `GetTopBottomPadding`:
https://github.com/ForNeVeR/xaml-math/blob/77d0bceeab5e6790cfb0b9a193b65f22b4217385/src/XamlMath.Shared/Atoms/MatrixAtom.cs#L139-L145

If one replace the `shift` argument by `0`:
``` csharp
Tuple<Box, Box> GetTopBottomPadding(int i, int j)
{
    var value = matrixCellGaps[i][j].Vertical;
    var topBox = new StrutBox(0.0, VerticalPadding / 2 + value, 0.0, 0.0);
    var bottomBox = new StrutBox(0.0, VerticalPadding / 2 + value, 0.0, 0.0);
    return new Tuple<Box, Box>(topBox, bottomBox);
}
```

The issue disappears:
image

I don't have enough hindsight to know if there are side effects.

Orace:
This may fix #387


ForNeVeR (approved):
Thanks!

Please let me know if the issue is critical for you, and you'd like to see a bugfix release with it.

PR #390, opened 2023-04-24
(#389) Fix padding issue with `\cases`.

Orace:
This fix the example from #389.

Ex: \cases{C &= \cases{ A & B \\ A & B} \\ M &= x\matrix{A & B \\ A & B}}
image

PR #392, opened 2023-04-26
Improve matrix atom

Orace:
WIP:

- [x] Depend on #390


ForNeVeR (approved):
Thanks, this is so great.

PR #394, opened 2023-04-26
WIP: Improve WpfMath.Example layout

Orace:
I'll be AFK for a week, hope you will like this PR.

Screenshot


ForNeVeR (changes requested):
I quite like the design change, but I don't think that the formula list should go into the menu.

Let's maybe try placing it in a combo-box above the formula entry field? What do you think?

When I test the examples, it's very useful to have them handy and not hidden into the menu.

Orace:
I did some changes:

image

ForNeVeR (approved):
That's great, thank you!

PR #398, opened 2023-05-03
Fix selection highlighting in FormulaControl

### LINKED ISSUE #397: Selection highlighting is broken in WPF
Since 44d3bac selection highlighting is broken in the WPF `FormulaControl`.

Orace:
This should fix #397


src/WpfMath/Controls/FormulaControl.xaml.cs
```diff
@@ -175,8 +175,11 @@ private void Render()
 
             using (var drawingContext = visual.RenderOpen())
             {
-                texFormula.RenderTo(drawingContext, environment, Scale);
+                var renderer = new WpfElementRenderer(drawingContext, Scale);
+                renderer.RenderElement(formulaBox, 0, formulaBox.Height);
+                renderer.FinishRendering();
```
Orace, line 180: Like it was done for the `TexFormula` type, this code can be moved to an extension method for the `Box` type, but I'm not sure this is of any interest.

ForNeVeR: Meh, as long as it's only one place, that's fine to just write here.

ForNeVeR (approved):
Thanks!

Issue #349, opened 2023-01-28
Is it possible to write crossed-out math?


ForNeVeR:
### Discussed in https://github.com/ForNeVeR/wpf-math/discussions/348

<div type='discussions-op-text'>

<sup>Originally posted by **TheDIMONDK** January 28, 2023</sup>
# Hello! Is it possible to do as here:
![image](https://user-images.githubusercontent.com/59896217/215277146-824a5de0-67c8-459c-85cb-6d93fde03b1e.png)

## [Source](https://tex.stackexchange.com/questions/75525/how-to-write-crossed-out-math-in-latex)</div>

PR #402, opened 2023-05-22
Add support for \cancel and co.

Orace:
This commit address #349 and add support to:

- [x] `\cancel`
- [x] `\bcancel`
- [x] `\xcancel`
- [ ] `\cancelto`

Current output of `\cancel{Q} \\ \xcancel{69} \\ \frac{a\bcancel{b}}{\bcancel{b}}`:

cancel


src/XamlMath.Shared/Boxes/ZBox.cs
```diff
@@ -0,0 +1,38 @@
+using System.Linq;
+using XamlMath.Rendering;
+
+namespace XamlMath.Boxes
+{
+    public class ZBox : Box
```
ForNeVeR, line 6: Maybe `LayeredBox`? What do you think? `ZBox` sounds kinda technical and wasn't obvious to me (I had to think for several seconds about that).

src/XamlMath.Shared/Boxes/ZBox.cs
```diff
@@ -0,0 +1,38 @@
...
+        public override int GetLastFontId()
+        {
+            var fontId = TexFontUtilities.NoFontId;
+            foreach (var child in Children)
+            {
+                fontId = child.GetLastFontId();
+                if (fontId == TexFontUtilities.NoFontId)
+                    break;
+            }
+            return fontId;
```
ForNeVeR, line 35: Is there a meaning to this logic? Otherwise, maybe just return the last child's font id?

ForNeVeR (changes requested):
Great work! There's a couple of minor issues left, could you please take a look?

I should note that this is a breaking change technically, but a good one and for a good reason, so I approve that.

Orace:
@ForNeVeR, this is a draft work, I'm open to (and ask for) any advice on this.

src/XamlMath.Shared/Boxes/ZBox.cs
Orace: I don't even know what it's used for 🙄, I copied it from `HorizontalBox`.

src/XamlMath.Shared/Boxes/ZBox.cs
Orace: Fixed in e9e1e1e.

src/XamlMath.Shared/Boxes/ZBox.cs
ForNeVeR: Aha, okay. Let's keep it as it is for now.

ForNeVeR:
I think we'll leave `\cancelto` still open for grabs for now.

PR #405, opened 2023-05-24
Simplify HorizontalBox.Add logic.

### LINKED ISSUE #403: The logic in HorizontalBox.Add is broken
Actual code reads:

https://github.com/ForNeVeR/xaml-math/blob/d4330c54e16ef1d30c484eabc1d21b22c6d1f505/src/XamlMath.Shared/Boxes/HorizontalBox.cs#L50-L59

But `Children.Count` is necessarily at least 1. Hence when the first child is added, the `Math.Max` first argument value is always 0 (initial value). This may cause an issue if the second argument is less than 0.

Orace:
This fix #403

### DIFF
--- a/src/XamlMath.Shared/Boxes/HorizontalBox.cs
+++ b/src/XamlMath.Shared/Boxes/HorizontalBox.cs
@@ -47,15 +47,15 @@ namespace XamlMath.Boxes
         {
         }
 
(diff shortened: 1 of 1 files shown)


ForNeVeR (approved):
Thanks!

PR #430, opened 2023-08-07
Refactored `TexPredefinedFormulaParser` to improve readability and ascertaining more invariants

Lehonti:
- Sealed the class (`TexPredefinedFormulaParser`) itself
- Fields with type `IDictionary<,>` never change (are only read, not written to) after the object is constructed, so I made them of type `IReadOnlyDictionary<,>`. All entry additions happen inside the constructor/s (I'm including the static constructor, as we will see in the next point).
- `typeMappings` and `argValueParsers` are always the same, in all instances, so I made them `static` and moved their construction to the static constructor. This can help us save on heap allocations.
- Sealed the `private` parser classes. Also, since most of them don't hold any internal state (and if the maintainer agrees) they can be turned into singletons in a future pull request.
- Nesting levels were decreased in the `Parse()` method through inversion of `if` blocks.

### DIFF
--- a/src/XamlMath.Shared/TexPredefinedFormulaParser.cs
+++ b/src/XamlMath.Shared/TexPredefinedFormulaParser.cs
@@ -11,13 +11,14 @@ using XamlMath.Utils;
 namespace XamlMath
 {
     // Parses definitions of predefined formulas from XML file.
-    internal class TexPredefinedFormulaParser
+    internal sealed class TexPredefinedFormulaParser
     {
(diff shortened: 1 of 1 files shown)


ForNeVeR (approved):
Yep, looks good!

ForNeVeR:
> Sealed the private parser classes. Also, since most of them don't hold any internal state (and if the maintainer agrees) they can be turned into singletons in a future pull request.

Feel free to do it in a separate PR.

PR #431, opened 2023-08-07
Sealed exception types in `XamlMath.Shared\Exceptions`

Lehonti:
Also, modernized namespace references (they are now file-scoped).

P.S. I don't know what is causing the build errors. Does sealing exception classes do anything that may cause them?


ForNeVeR:
> P.S. I don't know what is causing the build errors. Does sealing exception classes do anything that may cause them?

By sealing the classes, you've formally broken code compatibility: before that, the user code could inherit from these classes, but now it cannot.

I sincerely hope nobody actually inherited from them, so I believe we can safely merge this without bumping the major version 😅

ForNeVeR (approved):
Thank you!

PR #432, opened 2023-08-08
Private fields which are never modified were made `readonly`

Lehonti:
Done automatically by IDE, but occurrences were few, so reviewing it shouldn't be too daunting


ForNeVeR (approved):
Thanks, great!

PR #433, opened 2023-08-08
`SVGConverter.cs`: converted `if` towers to `switch` statements

Lehonti:
Specifically, these `if` towers were for pattern matching

Also, started using file-scoped namespace


ForNeVeR (approved):
A nice improvement, thank you!

PR #434, opened 2023-08-09
Internal refactoring - little nudge towards immutable code

Lehonti:
This is a nudge towards immutability, which gives us greater piece of mind when coding.

In `XamlMath.Shared` there are methods that create mappings and lists to be looked up. These methods usually return arrays or lists that expose the `IList<T>` interface.

Where possible, the methods were made to return `IReadOnlyList<T>` or `IEnumerable<T>` instead. This makes their intent clearer (they are not to be modified, just used for lookup).

Similarly, some methods that return `IDictionary<TKey,TValue>` were made to return `IReadOnlyDictionary<TKey,TValue>`.

The types of the fields that store these lists or mappings were changed accordingly.

`internal` classes involved were `sealed`.

No changes to the public API.


ForNeVeR (approved):
Yup, that's good!

PR #435, opened 2023-08-10
Extracted calculations in `HtmlColorParser.ParseSingleComponent` into independent methods

Lehonti:
I think it's more readable this way


ForNeVeR (approved):
Yep, looks better than before. Thanks!

PR #436, opened 2023-08-10
Sealed classes that _only_ have `private` constructor/s

Lehonti:
And I know that I formally broke API compatibility again :) but there's no way to inherit from these classes if the constructor is `private`


ForNeVeR (approved):
Thanks!

PR #437, opened 2023-08-12
Extracted method in `GeometryElementRenderer`

Lehonti:
Now you can apply a single transformation _or_ multiple transformations


ForNeVeR (approved):
Thanks!

PR #440, opened 2023-08-12
Moved initialization of typeface in `WpfSystemFont` into its own method, so that constructor is more concise

Lehonti:


ForNeVeR (approved):
Thanks!

Issue #29, opened 2017-02-19
Remove usage of ref parameters from the parser


ForNeVeR:
Current parser is messy a bit:  it uses `ref` parameters all the way. We need to refactor it and replace that with some alternate architecture that'll be easier to follow.

PR #442, opened 2023-08-12
Extracted method for parsing left group char in `TexFormulaParser`

Lehonti:
I see that this parser is mutation-based...In the future I'd like to make it more immutable, more functional, and more generation-based. This is a step in that direction (although I don't think that's obvious right now). Thankfully all methods are private so refactoring is easier

### DIFF
--- a/src/XamlMath.Shared/TexFormulaParser.cs
+++ b/src/XamlMath.Shared/TexFormulaParser.cs
@@ -248,16 +248,13 @@ namespace XamlMath
                 }
                 else if (ch == leftGroupChar)
                 {
-                    var groupValue = ReadElement(value, ref position);
-                    var parsedGroup = Parse(groupValue, textStyle, environment.CreateChildEnvironment());
-                    var innerGroupAtom = parsedGroup.RootAtom ?? new RowAtom(groupValue);
-                    var groupAtom = new TypedAtom(
-                        innerGroupAtom.Source,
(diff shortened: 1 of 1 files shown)


ForNeVeR:
> I see that this parser is mutation-based...In the future I'd like to make it more immutable, more functional, and more generation-based.

We even have an issue about that: #29

src/XamlMath.Shared/TexFormulaParser.cs
```diff
@@ -296,6 +293,20 @@ internal TexFormula Parse(SourceSpan value, string? textStyle, ICommandEnvironme
             return formula;
         }
 
+        private void ProcessLeftGroupChar(SourceSpan value, ref int position, string textStyle, ICommandEnvironment environment, TexFormula formula, int initialPosition)
```
ForNeVeR, line 296:
```suggestion
        private void ProcessLeftGroupChar(SourceSpan value, ref int position, string? textStyle, ICommandEnvironment environment, TexFormula formula, int initialPosition)
```

Remember the compiler warnings and that `textStyle` is nullable.

ForNeVeR (changes requested):
Other than a minor issue, LGTM!

Lehonti:
No problem, I accepted your suggestions :)

ForNeVeR (approved):
Thanks!

PR #443, opened 2023-08-12
Extracted box processing method in `FormulaControl`

Lehonti:
This already helps us decrease some nesting levels (see `ProcessBox()` method), but it should get a lot better after several refactorings


ForNeVeR (approved):
Thanks!

PR #445, opened 2023-08-26
Made singleton `Instance`s `readonly`

Lehonti:
I know that this formally breaks the API, but it doesn't make sense to re-assign those static fields, and also they don't have any internal state, so nothing should break


ForNeVeR (approved):
Thanks!

PR #446, opened 2023-08-26
Sealed more exception classes

Lehonti:
Also, in the particular case of these exceptions, not only does it not make sense to inherit from them, but their constructors are `internal`, so there already wasn't any way of inheriting from them


ForNeVeR (approved):
Thanks!

PR #447, opened 2023-08-26
`sealed` a few `internal` classes inheriting from `Box` and `Atom` in `XamlMath.Shared`

Lehonti:
They are not part of the public interface, and no class is inheriting from them, so they should be fine to seal


ForNeVeR (approved):
Thanks!

PR #457, opened 2023-09-05
`WpfMath.Converters.SVGConverter` now enumerates resulting strings with `yield return` instead of adding to `StringBuilder`

Lehonti:
Also, I'm not sure why `m_nestedLevel` is needed, so I commented it out, but the code is still there if it's needed

### DIFF
--- a/src/WpfMath/Converters/SVGConverter.cs
+++ b/src/WpfMath/Converters/SVGConverter.cs
@@ -1,49 +1,50 @@
 using System;
+using System.Collections.Generic;
 using System.Diagnostics;
 using System.Globalization;
-using System.Text;
 using System.Windows.Media;
 
 namespace WpfMath.Converters;
 
 public class SVGConverter
 {
-    private int m_nestedLevel = 0;
+    //private int m_nestedLevel = 0;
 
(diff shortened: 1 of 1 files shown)


src/WpfMath/Converters/SVGConverter.cs
```diff
@@ -1,49 +1,50 @@
...
 using System.Globalization;
-using System.Text;
 using System.Windows.Media;
 
 namespace WpfMath.Converters;
 
 public class SVGConverter
 {
-    private int m_nestedLevel = 0;
+    //private int m_nestedLevel = 0;
```
ForNeVeR, line 11: Meh, it was never used, let's just drop it.

ForNeVeR (changes requested):
A bit of minor issues, everything else LGTM!

src/WpfMath/Converters/SVGConverter.cs
Lehonti: Perfect, I just dropped it :) now let's just wait for the builds to finish :)

ForNeVeR (approved):
Great, thanks!

PR #461, opened 2023-09-26
Shortened `new()` expressions in several places

Lehonti:
The purpose is twofold: being more concise and reduce repetition.

Also, improved `null` checks in `Equals` methods of `SourceSpan`


ForNeVeR (approved):
Thanks!

PR #462, opened 2023-09-26
Removed unnecessary dictionary lookup

Lehonti:
Notice that the exact same value is being fetched just a line below that


Lehonti:
I will do this change in another pull request

PR #463, opened 2023-09-26
Converted `string.Format` to interpolated string

Lehonti:
Also, removed unnecessary dictionary lookup in `DefaultTexFontParser.cs`, in `GetDefaultTextStyleMappings()`. Notice that the variable just below the one that was deleted accesses the exact same value.


ForNeVeR (approved):
Thanks!

PR #464, opened 2023-09-27
In `TexPredefinedFormulaSettingsParser`, methods `GetSymbolMappings` and `GetDelimiterMappings` were refactored and they now share the same mechanism.

Lehonti:
The ugly `AddToMap()` method was removed and it was replaced with a generative approach


ForNeVeR (approved):
Thanks!

PR #466, opened 2023-09-27
Inside `MainViewModel.Export()`, exception handling was made using a more specific exception type

Lehonti:
This helps us get rid of the awkward variable assignment inside the `try` block, and we can just assign the variable directly

### DIFF
--- a/src/WpfMath.Example/MainViewModel.cs
+++ b/src/WpfMath.Example/MainViewModel.cs
@@ -110,14 +110,14 @@ public sealed class MainViewModel : INotifyPropertyChanged
 
     private void Export()
     {
-        // Choose file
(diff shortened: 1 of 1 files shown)


ForNeVeR (changes requested):
I recommend considering updating this approach.

Before this, any exceptions other than `TexParseException` would be caught by our handler, but now they aren't.

I believe it's a good practice to call everything (since we know exceptions of other types may be thrown), both for the purposes of example and for practical purposes (for easier debug in example app).

Let's consider specific handling for `TexParseException` while still keeping general exception handling as well?

For example, we may just add a simple annotation like `var message = ex is TexParseException ? $"Parse exception: {ex.Message}" : ex.Message`.

What do you think?

Lehonti:
Sure, we can do that. See if you like the updated version better

ForNeVeR (approved):
Thanks! Yep, I like this better.

PR #468, opened 2023-09-28
Split `BigOperatorAtom.CreateBoxCore` into smaller methods

Lehonti:
I think the original method is long and chaotic. Hopefully this can help improve it, even if it's not yet perfect

### DIFF
--- a/src/XamlMath.Shared/Atoms/BigOperatorAtom.cs
+++ b/src/XamlMath.Shared/Atoms/BigOperatorAtom.cs
@@ -1,5 +1,6 @@
 using System;
 using XamlMath.Boxes;
+using XamlMath.Fonts;
 
 namespace XamlMath.Atoms;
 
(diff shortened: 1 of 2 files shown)


src/XamlMath.Shared/Atoms/BigOperatorAtom.cs
```diff
@@ -23,91 +24,130 @@ private static Box ChangeWidth(Box box, double maxWidth)
...
+    }
+
+    private readonly record struct BoxForBaseAtom(Box BaseBox, double Delta)
+    {
+        internal BoxForBaseAtom AdjustedToMaxWidth(double maxWidth) =>
+            this with { BaseBox = ChangeWidth(BaseBox, maxWidth) };
+    }
+    private BoxForBaseAtom CreateBoxForBaseAtom(TexEnvironment environment)
+    {
+        if (BaseAtom is not SymbolAtom symbolAtom || BaseAtom.Type != TexAtomType.BigOperator)
```
ForNeVeR, line 126: _(Minor)_ I'd prefer these branches to be reversed. It is kinda hard to read the inverse condition, while the original "what to do if the base atom is a symbol atom and has a type of `BigOperator`) has lower cognitive load IMO.

What do you think?

ForNeVeR (changes requested):
Thank you, it indeed became better this way! So far, I only have a minor thing to discuss.

src/XamlMath.Shared/Atoms/BigOperatorAtom.cs
Lehonti: I must have done that as an intermediate step in some refactoring and then forgotten about it :) but yeah, I agree, inverted logic (i.e. `is not` and `!=`) may be harder to understand. I've now applied your suggestion

ForNeVeR (approved):
Thanks!

PR #469, opened 2023-09-28
Extracted some logic from `FractionAtom.CreateBoxCore` into its own methods

Lehonti:
I think the original method is long and chaotic. Hopefully this can help improve it, even if it's not yet perfect


ForNeVeR (approved):
Thanks!

PR #471, opened 2023-10-04
Extracted calculation of values to be rendered from `OverUnderBox.CreateRenderChoices` into its own method, for a more generative approach

Lehonti:


ForNeVeR (approved):
Thanks!

Issue #140, opened 2018-07-15
\text in subscripts should have smaller font


ForNeVeR:
Consider this sample:

![default](https://user-images.githubusercontent.com/92793/42731744-5e3dca1a-881c-11e8-89b9-b3f6ae1ebc7a.png)

The text `ЭП` should have smaller font size than `КБ`.

B3zaleel:
Have you tried using the FontSize Atom for the (Sub / Super)script pair? The Fontsize of the Font Size Atom should be calculated by multiplying the script factor by the Font Size of the current environment.

PR #483, opened 2024-03-08
Fixed #412

### LINKED ISSUE #412: Block height for `\text`
Consider this example:
```
\text{f(x)} = \cases{\text{1/3} & \text{if 0 ≤ x ≤ 1;} \\
\text{2/3} & \text{if 3 ≤ x ≤ 4;} \\
\text{0} & \text{elsewhere.}}
```
image

You may note that the text blocks are looking a off. Compare it with same block without `\text`, check the text block vertical offsets:
```
\text{f(x)} = \cases{1/3 & if\thickspace0 \leqslant x \leqslant 1; \\
2/3 & if\thickspace3 \leqslant x \leqslant 4; \\
0 & \text{elsewhere.}}
```
image

(I have specially left `elsewhere` in the `\text` block to compare the offset.)

Something's wrong here.

Thor181:
Closes #412.

### DIFF
--- a/src/WpfMath/Fonts/WpfSystemFont.cs
+++ b/src/WpfMath/Fonts/WpfSystemFont.cs
@@ -132,6 +132,18 @@ internal sealed class WpfSystemFont : ITeXFont
             , 1
 #endif
         );
-        return new TeXFontMetrics(formattedText.Width, formattedText.Height, 0.0, formattedText.Width, 1.0);
+        
+        var actualHeight = CalculateActualHeight(formattedText);
+
+        return new TeXFontMetrics(formattedText.Width, actualHeight, 0.0, formattedText.Width, 1.0);
+    }
+
+    private static double CalculateActualHeight(FormattedText formattedText)
+    {
+        var pointText = new Point(0, 0);
+        var geometry = formattedText.BuildGeometry(pointText);
+        var actualHeight = geometry.Bounds.Height;
+
+        return actualHeight;
     }
 }
--- a/src/WpfMath/Fonts/WpfSystemFont.cs
+++ b/src/WpfMath/Fonts/WpfSystemFont.cs
@@ -132,18 +132,29 @@ internal sealed class WpfSystemFont : ITeXFont
             , 1
 #endif
         );
-        
-        var actualHeight = CalculateActualHeight(formattedText);
 
-        return new TeXFontMetrics(formattedText.Width, actualHeight, 0.0, formattedText.Width, 1.0);
-    }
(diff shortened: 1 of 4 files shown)


ForNeVeR (changes requested):
In its current form, this PR introduces a very rough workaround for the actual problem (literally a workaround: replace the text of particular height with another height).

It is not useless since it shows a place where the fix may be applied, but I of course cannot merge that as-is. A more thorough investigation would be required.

Have you figured out what exactly is wrong with our metrics? The initial issue I'd summarize as "something's wrong with how we calculate the text baseline", but where's the actual baseline calculation in our code?

ForNeVeR:
(also, feel free to ignore the test failure; I won't ask you to fix it right now since this PR is more of a discussion starter than an applicable fix — which is totally okay)

Thor181:
You are right, the previous solution was very rough. There is a more elegant option. FormattedText calculates the height itself with extra padding. In this commit I converted the FormattedText to geometry and took its actual height.

Thor181:
Also I think this solves problem #140

ForNeVeR:
Alright. I have read through the code and tried to create something that will work, based on the WPF API:
```csharp
    private static TeXFontMetrics GetFontMetrics(char c, Typeface typeface)
    {
        var formattedText = new FormattedText(c.ToString(),
            CultureInfo.CurrentUICulture,
            FlowDirection.LeftToRight,
            typeface,
            1.0,
            Brushes.Black
#if !NET452
            , 1
#endif
        );
        formattedText.LineHeight = 1.0;

        var depth = formattedText.Height - formattedText.Baseline;
        return new TeXFontMetrics(formattedText.Width, formattedText.Baseline, depth, formattedText.Width, 1.0);
    }
```

My way should _theoretically_ work correctly as well: we base our character's parameters on the font _baseline_ and count the extruding element (if any) as _depth_.

In reality, however, it turns out to be less precise than your suggestion.

I have also tried combining some more obscure properties such as `formattedSize.Extent` and `formattedText.OverhangAfter`, with no avail: the actual geometry calculation is _just better_.

Also, some property values just doesn't make any sense: for example, `Extent` of the `f` character is considered to be `1.01`, more than the line height itself; most letters also give negative `OverhangAfter`, while I expect them all to give zero.

ForNeVeR:
Alright, I was able to get it right, while keeping the text baseline calculation. Will open a follow up issue though.

ForNeVeR (approved):
Thanks for help! Your pull request indeed has lead us to an interesting discussion, fix of the initially reported issue, and opening of a new one: #484.

Issue #484, opened 2024-03-09
Matrix should align its items with the baseline, not the total box


ForNeVeR:
Consider this formula:
```tex
\text{f(x)} = \cases{\text{1/3} & \text{if 0 ≤ x ≤ 1;} \\
\text{2/3} & \text{if 3 ≤ x ≤ 4;} \\
0..... & \colorbox{red}{\text{.elsewherge.}} \\
0..... & \colorbox{red}{\text{..}}}
```

Check how it looks currently:
![image](https://github.com/ForNeVeR/xaml-math/assets/92793/3b1d5ecf-64c1-4d57-80f8-c09e65779ec3)

As you can see, on line `elsewherge` the red box is high and the item itself is shifted up because of this.

On the line `..`, on contrary, the box is quite low and thus the item is also placed incorrectly.

Now, consider this patch that temporarily disables the character depth calculation for the system fonts:
```diff
Index: src/WpfMath/Fonts/WpfSystemFont.cs
IDEA additional info:
Subsystem: com.intellij.openapi.diff.impl.patch.CharsetEP
<+>UTF-8
===================================================================
diff --git a/src/WpfMath/Fonts/WpfSystemFont.cs b/src/WpfMath/Fonts/WpfSystemFont.cs
--- a/src/WpfMath/Fonts/WpfSystemFont.cs	(revision Staged)
+++ b/src/WpfMath/Fonts/WpfSystemFont.cs	(date 1710012618138)
@@ -155,6 +155,6 @@
         var scalingFactor = typeface.FontFamily.LineSpacing / typeface.FontFamily.Baseline;
         height /= scalingFactor;
 
-        return new TeXFontMetrics(formattedText.Width, height, depth, 0, 1.0);
+        return new TeXFontMetrics(formattedText.Width, height, 0, 0, 1.0);
     }
 }
```

After it, the `elsewherge` placement is nearly perfect:
![image](https://github.com/ForNeVeR/xaml-math/assets/92793/03a5b671-0250-4fd1-b4f6-e4136f9d2e1a)

So, the problem is that the characters properly report its baseline to the box model, but it cannot take it into account when placing the matrix items (I believe that we should just take `Height` instead of `TotalHeight` somewhere in the matrix calculation).

PR #485, opened 2024-03-12
Suggested solution #484

Thor181:


src/WpfMath/Fonts/WpfSystemFont.cs
```diff
@@ -155,6 +155,6 @@ private static TeXFontMetrics GetFontMetrics(char c, Typeface typeface)
         var scalingFactor = typeface.FontFamily.LineSpacing / typeface.FontFamily.Baseline;
         height /= scalingFactor;
 
-        return new TeXFontMetrics(formattedText.Width, height, depth, 0, 1.0);
```
ForNeVeR, line 158: Why have you reverted this? It seems wrong: it will stop preserving the "depth" of the system font characters. Example:
image
(see how the selection box is wrong now)

ForNeVeR (changes requested):

src/WpfMath/Fonts/WpfSystemFont.cs
Thor181: Your solution doesn't solve the problem. In addition, it breaks other types of matrices.
image
image


> Why have you reverted this? It seems wrong: it will stop preserving the "depth" of the system font characters.

Because the letters overhang is under the baseline. If an alignment by the baseline, then the overhang must be below the baseline (seems logic). Otherwise a characters will be above the required line (in this case will be need substract depth at boxes placing - it is also seems logic 🤔).

PR #489, opened 2024-12-02
A few methods in `TexFormulaParser` were modified so that they return the updated positions instead of mutating them

Lehonti:
These methods were moved to the bottom of the file to keep track of them

This is a step towards what was discussed in #442


ForNeVeR (approved):
Thanks!

PR #518, opened 2025-02-06
`TexFormulaParser.ReadElement` was modified so that it returns the updated position instead of mutating it

Lehonti:
Similar to #489, but the number of changes is enough to need its own pull request.

The end goal is having the `TexFormulaParser` returning the new values instead of mutating the existing ones.


ForNeVeR (approved):
Thanks!

PR #614, opened 2026-05-27
Prototype implementation for WinUI 3 with Win2D

SuGar0218:
image

Associated issue: https://github.com/ForNeVeR/xaml-math/issues/423

Current implementation is incomplete. Highlighting selected range and rendering image are not implemented.

There is still a issue: when loading font from file, long paths will cause UNABLE_RO_MASK_PATH, see https://github.com/microsoft/Win2D/issues/1025 .


SuGar0218:
Added support for selection.
image

ForNeVeR:
Don't worry about non-Windows CI failures too much for now, we'll find a workaround.

SuGar0218:
There is an error in nuget-push:

 > Could not read the Windows SDK's Platform.xml at C:\Program Files (x86)\Windows Kits\10\Platforms\UAP\10.0.22621.0\Platform.xml [D:\a\xaml-math\xaml-math\src\WinUIMath.Example\WinUIMath.Example.csproj]

Does it due to the missing Windows 11 SDK 10.0.22621.0 ?

ForNeVeR:
It is possible. Any idea what are the requirements for this to work? I know of the following options:
- nuget package dependencies
- some SDK workloads

If it is a workload requirement (quite possible), then we'd just need to update our CI to install the correct workload.

SuGar0218:
Try to install "Windows 11 SDK (10.0.22621.0)" in Visual Studio Installer?
image

ForNeVeR:
Not something I have any idea how to do on CI, unfortunately. The list of components available there out of the box is [public](https://github.com/actions/runner-images/blob/main/images/windows/Windows2025-Readme.md).

Perhaps there still is a corresponding workload?

I tried `dotnet workload search winui` — nothing.

`dotnet workload search maui` yields some results, though.

Unfortunately, I am not sure about relation between MAUI and WinUI, and I was unable to find anything quickly — so far. I'll investigate more.

SuGar0218:
I know nothing about CI. Here is what I found:

[Installed Windows SDKs in Windows Server 2025](https://github.com/actions/runner-images/blob/main/images/windows/Windows2025-Readme.md#installed-windows-sdks):

10.0.26100.0

[Installed Windows SDKs in Windows Server 2022](https://github.com/actions/runner-images/blob/main/images/windows/Windows2022-Readme.md#installed-windows-sdks):

10.0.17763.0
10.0.19041.0
10.0.22621.0
10.0.26100.0

If we are using Windows Server 2025, I try to set target framework to 26100 version?

ForNeVeR:
That's a good clue! Let's try switching to 26100 and see if it fixes the issue on `nuget-push`.

It seems that we accidentally have different Windows versions for test CI and publish CI — this is incorrect and I'll fix it in due time. For now, let's target the `windows-2025` image, the newer one.

SuGar0218:
For non-Windows CI failures, maybe we can add this to csproj files to deal with:
``` xml
<EnableWindowsTargeting>true</EnableWindowsTargeting>
```
which also exists in WpfMath.csproj.

